PR #4, opened 2020-09-01
Add DeepCopy kernel for DML device

PatriceVignola:


tensorflow/core/kernels/dml_deepcopy_op.cc
```diff
@@ -0,0 +1,80 @@
...
+    DmlDevice* device = static_cast<DmlDevice*>(ctx->device());
+    const auto& execution_context = device->GetExecutionContext();
+
+    D3D12BufferRegion input_buffer =
+        dml_util::CreateBufferForTensor(device, input);
+
+    D3D12BufferRegion output_buffer =
+        dml_util::CreateBufferForTensor(device, *output);
+
+    std::array<D3D12_RESOURCE_BARRIER, 2> barriers = {
```
jstoecker, line 45: I believe this may be a race condition if two deep copies occur on the same input tensor. To make this safe, the transitions and copies should occur inside the same lock.

adtsai: This is actually safe, as the lifetime of the buffer is tied to the `D3D12BufferRegion`. While this method holds those objects, nobody else can re-use those buffers.

---
In reply to: [481370801](https://github.com/microsoft/tensorflow-directml/pull/4#discussion_r481370801) [](ancestors = 481370801)

adtsai (approved):
:shipit:

jstoecker (approved):

PR #11, opened 2020-09-02
Add Pad+Conv2D fusions

PatriceVignola:
It's possible to merge a Pad/PadV2 node with Conv2D/_FusedConv2D under the following conditions:

- The Pad/PadV2 node is the first input of the Conv2D/_FusedConv2D node
- The filter shape is statically known
- The padding value is 0 (in the case of PadV2)

When all of the above conditions are true, we can effectively remove the padding node and transfer its values into the `explicit_paddings` attribute of the Conv2D/_FusedConv2D node and adding it to the previously embedded padding values of Conv2D/_FusedConv2D (VALID, SAME or EXPLICIT).

This pattern can happen quite often when models are written naively and don't take advantage of Conv2D's "paddings" attribute.

### DIFF
--- a/tensorflow/core/grappler/optimizers/remapper.cc
+++ b/tensorflow/core/grappler/optimizers/remapper.cc
@@ -1141,9 +1141,6 @@ Status AddFusedContractionNode(RemapperContext* ctx,
   NodeDef fused_op;
   fused_op.set_name(bias_add.name());
   fused_op.set_device(contraction.device());
-  fused_op.add_input(contraction.input(0));  // 0: input
-  fused_op.add_input(contraction.input(1));  // 1: filter
-  fused_op.add_input(bias_add.input(1));     // 2: bias
 
   if (IsConv2D(contraction)) {
     fused_op.set_op(kFusedConv2D);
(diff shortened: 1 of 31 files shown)


tensorflow/core/grappler/optimizers/remapper.cc
```diff
@@ -144,6 +149,15 @@ struct ContractionWithBatchNorm {
   float epsilon = 0.0;
 };
 
+// Pad node followed by a Conv2D.
+struct PadWithConv2D {
+  PadWithConv2D() = default;
+
+  int pad = kMissingIndex;
+  int conv_2d = kMissingIndex;
+  int32 new_padding_values[8];
```
adtsai: Should initialize this as well

tensorflow/core/grappler/optimizers/remapper.cc
```diff
@@ -565,6 +579,202 @@ bool FindConv2DWithBatchNormAndActivation(
...
 }
 
+bool FindPadWithConv2D(const RemapperContext& ctx, int node_index,
+                       PadWithConv2D* matched) {
+  const auto* conv_node_view = ctx.graph_view.GetNode(node_index);
+  const auto* conv_node_def = conv_node_view->node();
+  // Root of the pattern must be a Conv2D.
+  if (!IsConv2D(*conv_node_def)) return false;
+
+  // Check that Conv2D is in inference mode.
```
adtsai: Conv doesn't have an is_training mode, does it? (unlike batch_norm)

tensorflow/core/grappler/optimizers/remapper.cc
```diff
@@ -918,18 +1141,31 @@ Status AddFusedContractionNode(RemapperContext* ctx,
...
-  fused_op.add_input(bias_add.input(1));     // 2: bias
 
   if (IsConv2D(contraction)) {
     fused_op.set_op(kFusedConv2D);
     CopyConv2DAttributes(contraction, &fused_op);
+
+    // Optionally, look for a preceding "Pad" node that can be merged into
+    // Conv2D
+    PadWithConv2D pad_with_conv2d;
+    if (FindPadWithConv2D(*ctx, matched.contraction, &pad_with_conv2d)) {
```
adtsai, line 1168: Do we need a test in remapper_test.cc for this?

adtsai (approved):

tensorflow/core/grappler/optimizers/remapper.cc
PatriceVignola: True! Bad copy pasta :D

tensorflow/core/grappler/optimizers/remapper.cc
PatriceVignola: It can't hurt... But we don't even run the core tests anymore. I can add some tests and at least run them once, assuming we won't modify the remapper very often.

PR #13, opened 2020-09-02
Fix DML's GatherND dimension count

PatriceVignola:
DML currently requires all tensors to have the same dimension count, so we pad the indices and output dimensions to match the params dimension count.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # TensorFlow-DirectML <!-- omit in toc -->
 
+| PyPI Release                                                                                                      | Build (directml branch)                                                                                                                                                                                                                 |
+| ----------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


adtsai:
>         ctx->GetOutputDataType(0), flat_output_shape, flat_output_shape);

I believe you can also have the DmlTensorDesc do this for you, by supplying the `flat_params_shape` here. (non-blocking, just an observation)

---
Refers to: tensorflow/core/kernels/dml_gather_nd_op.cc:206 in f4e8620. [](commit_id = f4e8620a974760e82207fb51456fe12b6055e033, deletion_comment = False)

adtsai (approved):
:shipit:

jstoecker (approved):

PatriceVignola:
> >         ctx->GetOutputDataType(0), flat_output_shape, flat_output_shape);

I believe you can also have the DmlTensorDesc do this for you, by supplying the `flat_params_shape` here. (non-blocking, just an observation)

---
Refers to: tensorflow/core/kernels/dml_gather_nd_op.cc:206 in f4e8620. [](commit_id = f4e8620a974760e82207fb51456fe12b6055e033, deletion_comment = False) 

Hmm but only the last dimension should be the same as the params tensor. The other dimension should be the same as the first dimension of indices.

PR #15, opened 2020-09-02
Fix a couple of kernels that weren't handling empty tensors correctly

adtsai:
* ConvBackpropInput/Filter were attempting to zero their output tensor by reading garbage past the end of the input tensor, multiplying it by zero, and writing it to the output. Not only is it not necessarily valid to read past the end of the input tensor (as it may have been allocated near the end of the heap), multiplying by zero doesn't always give zero (as in the case of NaN). The kernel wrapper already handles no-ops (and correctly zeros the output tensors) so none of this is necessary.
* Correctly handle no-ops and empty tensors in reduce and dynamic stitch. The BFC allocator returns nullptr for zero-sized allocations, which means that it's legitimately invalid to try to create a buffer over an empty tensor (as its `data` pointer will be NULL).

This should fix the following tests:
* Conv2DTest.testConv2DBackpropFilterWithEmptyInput
* DynamicStitchTest.testZeroSizeTensor
* ExponentialOpTest.testEmpty
* LogarithmOpTest.testEmpty
* ProdReductionTest.testGradientWithZeros
* RaggedTensorToSparseOpTest.testBadSparseTensorRank
* RaggedTensorToSparseOpTest.testEmpty10

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_kernel_context.cc
+++ b/tensorflow/core/common_runtime/dml/dml_kernel_context.cc
@@ -249,8 +249,14 @@ DmlGpuEvent DmlKernelContext::ZeroBuffer(ID3D12Resource* dst, uint64_t offset,
       dst, offset, size_in_bytes, pattern);
 }
 
-DmlGpuEvent DmlKernelContext::ZeroBuffer(const D3D12BufferRegion& dst) const {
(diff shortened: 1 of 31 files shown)


PatriceVignola (approved):
LGTM

PR #16, opened 2020-09-06
Improve OOM error handling

PatriceVignola:
OOM errors can happen anytime when closing a command list because we batch commands together and close the command list when the number of commands reaches a threshold. OOM errors should never crash the process, but should instead bubble up and eventually throw a ResourceExhausted exception at the python level. This change adds a Status to the return type of functions that can cause an OOM error and removes a bunch of dead code.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_command_recorder.cc
+++ b/tensorflow/core/common_runtime/dml/dml_command_recorder.cc
@@ -302,54 +302,6 @@ void DmlCommandRecorder::FillBufferWithPattern(
   OnCommandRecorded();
 }
 
-void DmlCommandRecorder::ExecuteCommandList(
-    ID3D12GraphicsCommandList* command_list, _Outptr_ ID3D12Fence** fence,
-    _Out_ uint64_t* completion_value) {
-  if (!status_.ok()) return;
-
-  HRESULT hr = current_command_list_->Close();
-
-  if (dml_util::HrIsOutOfMemory(hr)) {
-    status_ = errors::ResourceExhausted("OOM when closing the command list");
-    return;
-  }
-
-  DML_CHECK_SUCCEEDED(hr);
-
-  if (operations_recorded_in_current_command_list_ != 0) {
-    pending_command_lists_.push_back(current_command_list_.Get());
-    pending_command_lists_cacheable_.push_back(true);
-  } else {
-    cached_command_lists_.push_back(current_command_list_.Get());
-  }
-
-  current_command_list_ = nullptr;
-  operations_recorded_in_current_command_list_ = 0;
-
-  pending_command_lists_.push_back(command_list);
-  pending_command_lists_cacheable_.push_back(false);
-
-  // Remember the descriptor heap and apply it to the next command list
-  auto heap = current_descriptor_heap_;
-  current_descriptor_heap_ = nullptr;
-  Open();
-
-  // The caller can re-use relevent resources after the next set of work to be
-  // flushed has completed.  Its command list hasn't been executed yet, just
-  // batched.
-  DmlGpuEvent gpu_event = queue_->GetNextCompletionEvent();
-  gpu_event.fence.CopyTo(fence);
-  *completion_value = gpu_event.fence_value;
-
-  // Trigger a flush of the command list, with the assumption that it contains
-  // enough GPU work that this will help parallelize GPU work with subsequent
-  // CPU work.
-  CloseAndExecute();
-  Open();
-
-  SetDescriptorHeap(heap);
-}
-
 void DmlCommandRecorder::ResourceBarrier(
     absl::Span<const D3D12_RESOURCE_BARRIER> barriers) {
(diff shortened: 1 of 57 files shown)


adtsai:
All these Status codes are error prone; there are a couple of spots you missed for example (like the calls to SetCommandRecorder). Instead of allowing these Status codes to spread "virally", would it be simpler to instead just store a piece of state on the command recorder that keeps track of whether an OOM condition has occurred? Then callers would only check the Status where it mattered, e.g. on Flush or CPU readback. The effects of (say) ResourceBarrier aren't observable until a flush happens anyway, so I'm not sure if there's a ton of value of returning a Status code at that point as opposed to just returning it on Flush.

tensorflow/core/common_runtime/dml/dml_command_recorder.cc
```diff
@@ -253,14 +254,16 @@ void DmlCommandRecorder::FillBufferWithPattern(
...
   current_command_list_->ResourceBarrier(1, &barrier);
+
+  return Status::OK();
 }
 
-void DmlCommandRecorder::ExecuteCommandList(
+Status DmlCommandRecorder::ExecuteCommandList(
     ID3D12GraphicsCommandList* command_list, _Outptr_ ID3D12Fence** fence,
     _Out_ uint64_t* completion_value) {
   DML_CHECK_SUCCEEDED(current_command_list_->Close());
```
adtsai: Missed this call to Close()

adtsai:
What about the creation of descriptor heaps, command allocators, and DML compiled operators? These all involve resources that are resident in GPU memory, so could also conceivably fail with E_OUTOFMEMORY if VRAM is exhausted. Do we need to apply to the same treatment to those objects too?

PatriceVignola:
> All these Status codes are error prone; there are a couple of spots you missed for example (like the calls to SetCommandRecorder). Instead of allowing these Status codes to spread "virally", would it be simpler to instead just store a piece of state on the command recorder that keeps track of whether an OOM condition has occurred? Then callers would only check the Status where it mattered, e.g. on Flush or CPU readback. The effects of (say) ResourceBarrier aren't observable until a flush happens anyway, so I'm not sure if there's a ton of value of returning a Status code at that point as opposed to just returning it on Flush.

I only added return status to functions that can trigger a flush (although I missed some). Since a call to ResouceBarrier can trigger a flush by incrementing the command count on the command list, we need a way to bubble it up to the caller so that we can gracely fail. Saving a status on the command recorder means that kernels would have to check for this status after calling any functions that can trigger a flush, but you would need to know which functions are those. If ResourceBarrier for example is a void function, how do you know that it can ultimately trigger a flush and that you should check for the status?

tensorflow/core/common_runtime/dml/dml_command_recorder.h
```diff
@@ -55,17 +56,17 @@ class DmlCommandRecorder {
...
   void ResourceBarrier(absl::Span<const D3D12_RESOURCE_BARRIER> barriers);
-  void AddUAVBarrier();
 
   void CloseAndExecute();
 
   // If false, there are no pending commands to be submitted which indicates
   // that CloseAndExecute() would be a no-op.
   bool HasUnflushedWork() const;
 
+  Status GetStatus() const { return status_; }
```
adtsai, line 63: Hmm, it seems like it'd be simpler to just return the Status from CloseAndExecute and ExecuteCommandList. That way you wouldn't need these Get/Reset methods.

tensorflow/core/kernels/dml_kernel_wrapper.h
```diff
@@ -143,8 +143,8 @@ class DmlKernelWrapper : public DmlKernelWrapperBase {
     return std::make_shared<const typename TKernel::InitHelper>(ctx, attr_);
   }
 
-  DmlGpuEvent ComputeKernel(DmlKernel* kernel,
-                            DmlKernelContext* context) const override {
+  StatusOr<DmlGpuEvent> ComputeKernel(
```
adtsai, line 146: Does this ever actually return a Status?

adtsai (approved):

tensorflow/core/kernels/dml_kernel_wrapper.h
PatriceVignola: A few times. Before you would sometimes need to delegate the validation to an awkward void function if you wanted to validate something in Compute, just to be able to call OP_REQUIRES(...). Now, it's cleaner to simply call TF_RETURN_IF_ERROR and it avoids all the awkward `if (!ctx->status().ok()) return;` calls.

tensorflow/core/common_runtime/dml/dml_command_recorder.h
PatriceVignola: We still need a way to reset the status when flushing to let the user try something else after catching the exception. Otherwise, the status will permanently be in a failed state. We can't reset it in CloseAndExecute() since it may be called multiple times before we get a chance to flush.

PR #26, opened 2020-09-14
Resurrect core tests

PatriceVignola:
- Resurrect the core tests by extracting the right DirectML DLL from the python package, renaming it to DirectML.dll and setting `TF_DIRECTML_PATH` to its location
- Onboard the remapper tests to make sure that the new fusions that we added in our branch don't break anything

### DIFF
--- a/third_party/dml/ci/test/DispatchTestJobs.ps1
+++ b/third_party/dml/ci/test/DispatchTestJobs.ps1
@@ -144,7 +144,7 @@ foreach ($AgentInfo in $AgentsInfo)
             'RunOnWsl' = $AgentInfo.RunOnWsl;
         }
 
-        Write-Host $Params
+        Write-Host $Params.Artifacts
 
         $Build = $Ado.QueuePipeline(
             $Pipeline.id, 
--- a/third_party/dml/ci/test/DispatchTestJobs.ps1
+++ b/third_party/dml/ci/test/DispatchTestJobs.ps1
@@ -144,6 +144,8 @@ foreach ($AgentInfo in $AgentsInfo)
             'RunOnWsl' = $AgentInfo.RunOnWsl;
         }
 
+        Write-Host $Params
+
         $Build = $Ado.QueuePipeline(
             $Pipeline.id, 
             $AgentInfo.AgentQueueId,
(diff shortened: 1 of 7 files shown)


third_party/dml/ci/test/DispatchTestJobs.ps1
```diff
@@ -47,36 +53,78 @@ $Ado = [ADOHelper]::CreateFromPipeline($AccessToken)
...
+    }
+
+    $AgentsInfo.Add($WslAgentInfo)
+}
+
+$DispatchedJobs = [System.Collections.ArrayList]::new()
+
+foreach ($AgentInfo in $AgentsInfo)
 {
     $JobInfo =
```
jstoecker, line 105: Now that this script dispatches to multiple agent pools, it's possible to have jobs dispatched to two agents with the same name. This will almost certainly break CreateTestEmail.ps1, since it assumes unique agent names. One easy workaround might be embedding the agent pool in the name (e.g. `'AgentName' = "$($AgentInfo.TestPoolName):$($AgentInfo.Name)"`).

PatriceVignola: Oops, sorry about that. I've been having problems with the vscode extension where it doesn't always use the current branch to make the PRs, and I didn't double-check here. I'll re-create this PR but I'll keep your comments in mind for my other branch.

third_party/dml/ci/test/TestAgent.ps1
```diff
@@ -34,9 +43,28 @@ try
...
         $Results.Time.Start = (Get-Date).ToString()
-        py run_tests.py --test_group $TestGroup --tensorflow_wheel $TensorFlowWheelPath | Out-File -FilePath "test_${TestGroup}_log.txt"
-        $Results.Time.End = (Get-Date).ToString()
 
+        if ($RunOnWsl)
+        {
+            Push-Location $WslArtifactFolderAsWin
+
+            $LoadLibraryPath = wsl echo $WslArtifactFolder`:`$LD_LIBRARY_PATH
+            $WslTensorFlowWheelPath = Get-ChildItem tensorflow_directml-*-linux_x86_64.whl | Select-Object -First 1 -ExpandProperty Name
```
jstoecker, line 52: At some point we really should be producing "manylinux" wheels, so let's make this glob a little more permissive to avoid headaches later.

PR #27, opened 2020-09-15
Resurrect core tests

PatriceVignola:
- Resurrect the core tests by extracting the right DirectML DLL from the python package, renaming it to DirectML.dll and setting `TF_DIRECTML_PATH` to its location
- Onboard the remapper tests to make sure that the new fusions that we added in our branch don't break anything

### DIFF
--- a/third_party/dml/ci/test/run_tests.py
+++ b/third_party/dml/ci/test/run_tests.py
@@ -81,11 +81,15 @@ def main():
 
           # Rename the DLL to DirectML.dll since TF_DIRECTML_PATH only works
           # with that name
-          dll_folder = f"{temp_dir}/_solib_directml"
-          old_dll_path = glob.glob(f"{dll_folder}/DirectML*.dll")[0]
-          new_dll_path = f"{dll_folder}/DirectML.dll"
-          os.rename(old_dll_path, new_dll_path)
-          conda_env.run(command_line, {"TF_DIRECTML_PATH": dll_folder})
+          lib_folder = f"{temp_dir}/_solib_directml"
+
+          old_glob = "DirectML*.dll" if os.name == "nt" else "libdirectml.so.*"
+          new_name = "DirectML.dll" if os.name == "nt" else "libdirectml.so"
+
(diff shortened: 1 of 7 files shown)


third_party/dml/ci/test/run_tests.py
```diff
@@ -69,7 +72,22 @@ def main():
         for flag in json_test_group["flags"]:
           command_line.append(flag)
 
-      conda_env.run(command_line)
+      if not is_python_test and args.tensorflow_wheel is not None:
+        # .whl files are simple zip files, so we can extract DirectML.dll from
+        # it
+        with TemporaryDirectory() as temp_dir:
+          with ZipFile(args.tensorflow_wheel, 'r') as zip_wheel:
+            zip_wheel.extractall(temp_dir)
```
jstoecker, line 80: Nit: the wheels can be slow to extract, so it might be better to just extract the solib_directml folder. Though really a drop in the bucket for total test time... feel free to WontFix.

third_party/dml/ci/test/run_tests.py
```diff
@@ -69,7 +72,22 @@ def main():
...
+
+          # Rename the DLL to DirectML.dll since TF_DIRECTML_PATH only works
+          # with that name
+          dll_folder = f"{temp_dir}/_solib_directml"
+          old_dll_path = glob.glob(f"{dll_folder}/DirectML*.dll")[0]
```
jstoecker: May as well do the same for WSL (libdirectml.so.SHA -> libdirectml.so).

jstoecker (approved):

PatriceVignola:
> Nit: the wheels can be slow to extract, so it might be better to just extract the solib_directml folder. Though really a drop in the bucket for total test time... feel free to WontFix. 

Eh, it's always been instant for me when doing it from python (as opposed to powershell or windows explorer :P)

PatriceVignola:
> May as well do the same for WSL (libdirectml.so.SHA -> libdirectml.so). 

Good catch.

PR #29, opened 2020-09-15
Remove dependency on -fms-extensions and clean up winadapter

jstoecker:
This change paves the way for Linux builds that don't rely on Microsoft-specific compiler extensions, which are only available in some toolchains (i.e. Clang). The upstream builds use GCC, and it should be possible to build tensorflow-directml with GCC as well.

A requirement for this change is a portable solution for handling COM interfaces, IIDs, and GUIDs. The DirectX Shader Compiler (also open source) handles this by extending interfaces with a [DECLARE_CROSS_PLATFORM_UUIDOF macro](https://github.com/microsoft/DirectXShaderCompiler/issues/1342); however, this solution would require invasive changes to DX headers. Instead, a function template `uuidof<T>` is declared (and deleted, so there is no default implementation) and specialized for the interfaces required in this project. To avoid unintentionally initializing IIDs with the wrong value, Windows builds will compare and statically asserts with IIDs in the Windows SDK.

One important implication of this change is that dxguids.h will need to be extended if a new DX interface dependency is added *and* instantiated (e.g. a hypothetical `ComPtr<ID3D12Device10>`).

- Remove dependency on -fms-extensions (__uuidof, __declspec(uuid(x)), __declspec(selectany), ...).
- Address new warnings-as-errors (narrowing conversions, template ambiguity) in a few DML kernels.
- Clean up redundant options (include dirs) in targets that depend on winadapter bazel target.
- Replace unnecessary usage of __uuidof(x) with explicit IIDs.
- Switch WSL linker dir from /mnt/c/Windows/System32/lxss/lib to drive-agnostic /usr/lib/wsl/lib path.
- Update d3d12sdklayers.h from a very old version.
- Add dxguids.h/cc for declaring and initializing DirectX GUID helpers that don't require compiler intrinsics.
- Remove several types/macros in winadapter.h that are not necessary (the fewer the better...).
- Update wrladapter.h to use portable uuidof<T> function template, which doesn't require compiler intrinsics.


adtsai (approved):
:shipit:

PR #30, opened 2020-09-16
Handle device hung removal

PatriceVignola:
Device hung errors should allow the user to fallback instead of crashing the process.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_command_recorder.cc
+++ b/tensorflow/core/common_runtime/dml/dml_command_recorder.cc
@@ -362,9 +362,20 @@ void DmlCommandRecorder::CloseAndExecute() {
   // opened.
   current_descriptor_heap_ = nullptr;
 
+  HRESULT dml_reason = dml_device_->GetDeviceRemovedReason();
+  HRESULT d3d12_reason = d3d_device_->GetDeviceRemovedReason();
+
+  if (dml_reason == DXGI_ERROR_DEVICE_HUNG ||
+      d3d12_reason == DXGI_ERROR_DEVICE_HUNG) {
+    status_ = errors::Unknown(
+        "Device hung while executing a command. This can happen when the GPU "
+        "times out or when there's a problem in the driver.");
+    return;
+  }
+
   // Fail early if something horrifying happens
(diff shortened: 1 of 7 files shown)


adtsai:
Devices are global singletons, so device removal isn't a recoverable error. Unless we build in a mechanism to re-create devices that have been removed, I don't think there's any advantage to trying to handle this "gracefully" because in case of device removal the only recourse is to terminate the process.

PatriceVignola:
> Devices are global singletons, so device removal isn't a recoverable error. Unless we build in a mechanism to re-create devices that have been removed, I don't think there's any advantage to trying to handle this "gracefully" because in case of device removal the only recourse is to terminate the process.

It's true that we can't recover the DML device, but it should be possible to use another device (maybe it should be made clearer that the DML device cannot be used anymore in the error message). For example, when getting OOM errors, it's common to catch the exception and fallback to the CPU device. The bottom line is that crashes really shouldn't happen often in TensorFlow, unless it's a programmer error (that's what all the CHECKS in the code are for). Not to mention that seeing `Check failed: (((HRESULT)((dml_device_->GetDeviceRemovedReason()))) >= 0) == true (0 vs. 1)` as an error message tells you nothing about what went wrong.

adtsai:
Ah, I hadn't considered the possibility of falling back to CPU. In that case is there any reason to single out DXGI_ERROR_DEVICE_HUNG? Presumably DRIVER_ERROR or any of the other device removal reasons are equally bad (and equally unrecoverable).

tensorflow/core/common_runtime/dml/dml_command_recorder.cc
```diff
@@ -362,9 +362,22 @@ void DmlCommandRecorder::CloseAndExecute() {
   // opened.
   current_descriptor_heap_ = nullptr;
 
+  HRESULT device_removed_reason = dml_device_->GetDeviceRemovedReason();
+
+  if (device_removed_reason == S_OK) {
```
adtsai: >device_removed_reason == S_OK [](start = 6, length = 29)

Use SUCCEEDED() here

adtsai (approved):
:shipit:

PR #44, opened 2020-09-18
User/xianz/topk

zhangxiang1993:
the new added kernels are verified by running [topk_op_test](https://github.com/microsoft/tensorflow-directml/blob/directml/tensorflow/python/kernel_tests/topk_op_test.py).

testTopKGradients test is temprorary disabled because [scatter_nd](https://www.tensorflow.org/api_docs/python/tf/scatter_nd) is not implemented.

testTop1AllNan is also temproray disabled.


tensorflow/core/kernels/dml_topk_op.cc
```diff
@@ -0,0 +1,145 @@
...
+    const auto& input_in = ctx->input(0);
+    OP_REQUIRES(ctx, input_in.dims() >= 1,
+                errors::InvalidArgument("input must be >= 1-D, got shape ",
+                                        input_in.shape().DebugString()));
+    OP_REQUIRES(ctx, input_in.dim_size(input_in.dims() - 1) >= k,
+                errors::InvalidArgument(
+                    "input must have at least k columns. Had ",
+                    input_in.dim_size(input_in.dims() - 1), ", needed ", k));
+  }
+  bool GetSorted() const { return attr_->sorted_; }
```
fdwr: `GetSorted` isn't actually used below.

zhangxiang1993: That's true, the order of output of topk(..., sorted=False) is not defined in tensorflow. The test will just verify the output set. I'll delete the methods to return sorted_.

---
In reply to: [491183198](https://github.com/microsoft/tensorflow-directml/pull/44#discussion_r491183198) [](ancestors = 491183198)

PatriceVignola:
Why is testTop1AllNan failing? What is the output of DirectML versus the output of CPU/CUDA? If we need to add a special scenario for NaN numbers, we should add it in the kernel instead of just disabling the test. Conformance should be more important than performance, especially for operators that are not executed that often in a model like TopK.

jstoecker:
Let's make sure PR titles and associated merge commits are more descriptive.

jstoecker:
> Why is testTop1AllNan failing? What is the output of DirectML versus the output of CPU/CUDA? If we need to add a special scenario for NaN numbers, we should add it in the kernel instead of just disabling the test. Conformance should be more important than performance, especially for operators that are not executed that often in a model like TopK.

I agree with Pat here. NaNs tend to be more common in training, so we should understand the behavior difference if there is one. We should only disable tests when we have a limitation that is by design or well understood (e.g. missing data type).

zhangxiang1993:
Should we file a DML bug, and keep this test enabled?

---
In reply to: [696428476](https://github.com/microsoft/tensorflow-directml/pull/44#issuecomment-696428476) [](ancestors = 696428476)

jstoecker:
> Should we file a DML bug, and keep this test enabled?

Looking at the git blame, this test was added to avoid AV since the CPU op left indices uninitialized in this particular case. Since DML always returns valid indices, this shouldn't be a problem for us; I'd skip the test on DML devices with a comment explaining this.

tensorflow/python/kernel_tests/topk_op_test.py
```diff
@@ -108,6 +108,10 @@ def testTop3(self):
     values = -np.sort(-inputs)[:k]
     self._validateTopK(inputs, k, values, indices)
 
+  # The output indices from TF are all zeros because they won't be set due to 
+  # NaN != NaN, while DML API always return valid indices that might not match TF 
+  # behavior, so this test is diabled because this issue doesn't affect the kernel.
```
PatriceVignola: nit: line length

tensorflow/python/kernel_tests/topk_op_test.py
```diff
@@ -208,6 +212,8 @@ def testKTooLarge(self):
       nn_ops.top_k(inputs, 4)
 
   @test_util.run_deprecated_v1
+  # scatter_nd is not implemented in DML.
+  @test_util.skip_dml
```
PatriceVignola: Don't disable the test because scatter_nd isn't implemented, especially if you don't link a bug for it. We have many failing tests because of scatter_nd, so it would be a pain to hunt all the disabled tests once we implement it. ScatterNd is also probably one of the next operators that I'm going to implement, so it won't fail for too long.

jstoecker (approved):
Looks good. Just make sure the merge isn't called "/User/xianz/topk" :)

PR #49, opened 2020-09-23
Fix default window sizes for StridedSlice

PatriceVignola:
When the input size and output size of a dimension are equal, the kernel logic tries to collapse window dimensions together in order to allow more dimensions to be supported. But when it does so, it doesn't overwrite the default values which are set to 0. This problem was previously hidden by the fact that DML didn't validate the range of the window sizes, and it just happened that the kernel worked when `WindowSize[i] == 0`.


adtsai (approved):
:shipit:

PR #51, opened 2020-09-24
Fix FuseConv2DWithSqueezeAndBias remapper test

PatriceVignola:
The test was using a temporary variable to initialize the Axis array, which was getting destroyed as soon as the function returned.


adtsai (approved):
:shipit:

PR #52, opened 2020-09-24
Update DirectMLX, part 2

adtsai:


tensorflow/core/common_runtime/dml/dml_util.cc
```diff
@@ -201,17 +201,40 @@ DmlTensorLayout GetDmlTensorLayout(TensorFormat format, uint32_t rank) {
   return tensor_layout;
 }
 
-dml::TensorLayout GetDmlXTensorLayout(TensorFormat format) {
+dml::TensorPolicy GetDmlXTensorPolicy(TensorFormat format) {
   switch (format) {
     case FORMAT_NHWC:
-      return dml::TensorLayout::Nhwc;
+      return dml::TensorPolicy::InterleavedChannel();
```
PatriceVignola, line 207: Tensorflow uses "channels_first" to name NCHW and "channels_last" to name NHWC. Personally, I think it's more descriptive than "InterleavedChannel", and it follows the existing tensorflow conventions.

third_party/dml/DirectMLX.h
```diff
@@ -247,79 +248,132 @@ namespace dml
...
+    // The custom properties returned by a TensorPolicy.
+    struct TensorProperties
     {
-        Default = 0,
-        Nchw = 0,
-        Nhwc = 1,
+        Optional<TensorDimensions> strides;
+        uint64_t totalTensorSizeInBytes;
+        uint32_t guaranteedBaseOffsetAlignment;
     };
```
PatriceVignola, line 259: Now that we have a strides policy, do we still need to have a "strides" parameter as part of the gather operators?

adtsai: Nope, which is why they've been removed. The int64 striding is now handled by a custom tensor policy (see dml_util.cc)

tensorflow/core/common_runtime/dml/dml_util.cc
adtsai: I've never much liked "channels first" (because they really aren't). "Interleaved" is used distinct from "Planar", which is terminology from the graphics world. "Default" and "ChannelsLast" might be reasonable, though. I'll have to think about it...

jstoecker (approved):

PR #55, opened 2020-09-26
Upgrade GRPC to support glibc 2.31

PatriceVignola:
This is inspired from this change: https://github.com/tensorflow/tensorflow/commit/f396035891b0938364ea247a7dd243a147930c6e

The previous GRPC doesn't work with glibc >= 2.30. Since glibc 2.31 comes by default with Ubuntu 20.04, which is the default Ubuntu distro in the Microsoft Store, it's a good idea to be able to build the project on it.


jstoecker:
Ah yes, I was aware of this issue but didn't get around to fixing it. Thank you!

jstoecker (approved):

PR #60, opened 2020-09-28
Add SoftsignGrad kernel for DML

zhangxiang1993:


PatriceVignola (approved):
:shipit:

jstoecker:
@zhangxiang1993 : I didn't see this in time, but it would be faster to remove the pow and do a multiply. Or, alternatively, we could make the DMLX helper for Pow a little smarter and apply this automatically when the exponent is 2. @adtsai for his opinion on DMLX.

PR #64, opened 2020-09-30
Release memory after a device has been removed

PatriceVignola:


tensorflow/core/common_runtime/dml/dml_command_recorder.cc
```diff
@@ -370,12 +375,9 @@ void DmlCommandRecorder::CloseAndExecute() {
...
-    status_ = errors::Unknown(
-        "Device was removed because of the following reason: ",
-        dml_util::StringifyDeviceRemovedReason(device_removed_reason),
-        ". This can happen when the GPU times out or when there's a problem in "
-        "the driver. You won't be able to use this DML device again in the "
-        "current process.");
+    // Signal to the listeners that the device has been removed, so they can
+    // handle it appropriately (e.g. release memory)
+    device_removed_event_->NotifyListeners();
     return;
```
adtsai, line 381: Introducing this kind of re-entrancy is often difficult to reason about, because you don't know what the state of the world is when this callback gets invoked. The command recorder is the one who decides when to invoke this, but how do the listeners know whether it's safe to do something (e.g. release all memory) without having to understand the implementation details of the command recorder?

Don't we already have a central place where we check for device removal anyway? (Presumably sometime around when we flush) Can we just respond to the device removal at that point?

PatriceVignola: Pooled heaps need to be notified of device removals to release the memory.

The listeners know it's "safe" to do something like releasing memory because it knows the state of the device is now removed, so nothing should try to use that memory anymore or hold on to it. We're not attempting to rescue the device here, but merely fail in a way that allows other devices to keep going.

tensorflow/core/common_runtime/dml/dml_pooled_heap.cc
```diff
@@ -196,6 +206,12 @@ void DmlPooledHeap::ReclaimAllocations() {
   }
 }
 
+void DmlPooledHeap::OnDeviceRemoved() {
+  chunks_.clear();
+  total_capacity_ = 0;
```
adtsai, line 211: If this method gets called from a callback, what guarantees thread-safety here? Both the readback heap and upload heap are meant to be thread-safe.

adtsai: (Similar comment with the DeviceRemoved() method - there's nothing to guarantee that device_removed_ isn't being written to at the same time DeviceRemoved() is trying to read from it)

---
In reply to: [498592247](https://github.com/microsoft/tensorflow-directml/pull/64#discussion_r498592247) [](ancestors = 498592247)

tensorflow/core/common_runtime/dml/dml_upload_heap.cc
```diff
@@ -53,7 +63,16 @@ StatusOr<DmlGpuEvent> DmlUploadHeap::BeginUploadToGpu(
   // Map the upload heap and copy the source data into it at the specified
   // offset
   void* upload_heap_data = nullptr;
-  DML_CHECK_SUCCEEDED(chunk->resource->Map(0, nullptr, &upload_heap_data));
+  HRESULT hr = chunk->resource->Map(0, nullptr, &upload_heap_data);
+
+  if (hr == DXGI_ERROR_DEVICE_REMOVED) {
+    HRESULT device_removed_reason = d3d12_device_->GetDeviceRemovedReason();
```
adtsai, line 69: This seems kinda fragile... for one, this only checks for one of the many possible device removal reasons (e.g. what if Map() returns DXGI_ERROR_DEVICE_HUNG?) But I wonder what the intended end game is here... since device removals are asynchronous, *any* call into D3D12 (that returns an HRESULT) could potentially return one of the device removed error codes. Are we intending to check every single call into D3D12 for device removal?

tensorflow/core/common_runtime/dml/dml_pooled_heap.cc
PatriceVignola: Well, it's called from the execution context's `Flush()`, which has a lock. Nothing stops external callers from calling `Trim()` from thread-unsafe code either, but the caller shouldn't do that.

`DeviceRemoved()` is a protected function that is only called under locks as well. There's no way to call it externally.

tensorflow/core/common_runtime/dml/dml_upload_heap.cc
PatriceVignola: I assume most of them will happen when executing a command list (given the large number of operations submitted at once), but yeah, we should ideally cover all the cases. Device removal isn't a user error and isn't a programming error from us, so it shouldn't just crash. It can be repro'ed as easily as detaching your surface book.

What are some of the cases that I missed?

tensorflow/core/common_runtime/dml/dml_pooled_heap.cc
adtsai: The execution context has a lock, but the pooled heap doesn't execute underneath that lock (the upload heap and readback heap maintain their own independent locks). What stops one thread from calling into the execution context and triggering a device removal, simultaneous with a thread calling into the upload/readback heap and modifying this internal state?

---
In reply to: [498945010](https://github.com/microsoft/tensorflow-directml/pull/64#discussion_r498945010) [](ancestors = 498945010)

PatriceVignola: Good point, I didn't think about this scenario. `OnDeviceRemoved()` should acquire the lock as well.

PR #72, opened 2020-10-19
Use NCHW for FusedBatchNormGrad

PatriceVignola:
Jeff noticed that operators used inside FusedBatchNormGrad (e.g. Reduce) didn't have a NCHW layout inside DML. This is because grappler explicitly uses NHWC by default for FusedBatchNormGrad when is_training == false. This is because the GPU kernel does layout transformations on a case-by-case basis since cuDNN has some NHWC fast paths. For DML on the other hand, NCHW should always be faster than NHWC at the moment.

### DIFF
--- a/tensorflow/core/grappler/optimizers/generic_layout_optimizer_transposer.cc
+++ b/tensorflow/core/grappler/optimizers/generic_layout_optimizer_transposer.cc
@@ -776,6 +776,10 @@ Status FusedBatchNormGradTransposer::TransposeNode(
 
   if (!IsTraining(*node)) {
     string task, device;
+
+    // Unlike the CUDA kernel, DML's FusedBatchNormGrad should always be faster
+    // with NCHW, so we transform the layout here instead of doing it
+    // in the kernel itself.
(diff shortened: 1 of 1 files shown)


tensorflow/core/grappler/optimizers/generic_layout_optimizer_transposer.cc
```diff
@@ -769,10 +769,22 @@ bool FusedBatchNormGradTransposer::IsTraining(
...
     TransposeContext* context, utils::MutableNodeView* node) {
   DCHECK(IsFusedBatchNormGrad(*node->node()));
-  if (!ShouldProcess(*context, *node) || !IsFanoutPortRankN(*node, 0, 4) ||
-      !IsTraining(*node)) {
+
+  if (!ShouldProcess(*context, *node) || !IsFanoutPortRankN(*node, 0, 4)) {
     return Status::OK();
   }
+
+  if (!IsTraining(*node)) {
```
adtsai, line 777: Would be worth adding a comment here explaining the reason for DML's special case.

adtsai (approved):

PR #77, opened 2020-10-29
Add TensorScatterUpdate, TensorScatterAdd and TensorScatterSub

PatriceVignola:


tensorflow/core/kernels/dml_tensor_scatter_ops.cc
```diff
@@ -0,0 +1,385 @@
...
+    result = BinaryOp()(input, result);
+
+    Microsoft::WRL::ComPtr<IDMLCompiledOperator> compiled_op =
+        scope.Compile(DML_EXECUTION_FLAG_NONE, {result});
+
+    Initialize(ctx, std::move(tensors), compiled_op.Get());
+  }
+
+  StatusOr<DmlGpuEvent> Compute(DmlKernelContext* ctx) const override {
+    D3D12BufferRegion input_buffer =
```
adtsai, line 298: Can we eliminate all this custom logic by using FillValueConstant to generate the initial empty scalar?

PatriceVignola: Sure we can. My thought process was that zeroing the buffer would be faster than adding another dispatch.

adtsai: Maybe we should keep a pre-zero'd tensor attached to the device for situations such as these - I think we have several other instances where we do stuff like this. It's kinda similar to how CPUs have a zero register. Either way it's not a big deal, so I'll leave the decision up to you.

adtsai (approved):
:shipit:

tensorflow/core/kernels/dml_tensor_scatter_ops.cc
```diff
@@ -0,0 +1,385 @@
...
+          .TypeConstraint<int64>("Tindices"),                               \
+      DmlKernelWrapper<DmlTensorScatterUpdateKernel<int64>,                 \
+                       GetOutputShapeAsInputShapeHelper>)                   \
+  REGISTER_KERNEL_BUILDER(                                                  \
+      Name("TensorScatterAdd")                                              \
+          .Device(DEVICE_DML)                                               \
+          .TypeConstraint<type>("T")                                        \
+          .TypeConstraint<int32>("Tindices"),                               \
+      DmlKernelWrapper<                                                     \
+          DmlTensorScatterBinaryKernel<int32, std::plus<dml::Expression>>,  \
```
jstoecker, line 355: It's cool that this works. I know why it works (DMLX overloads), but it's still cool. :)

jstoecker (approved):

PR #79, opened 2020-10-30
Add ParallelConcat, _ParallelConcatStart and _ParallelConcatUpdate

PatriceVignola:


tensorflow/core/kernels/inplace_ops.cc
```diff
@@ -250,6 +250,29 @@ REGISTER_KERNEL_BUILDER(Name("_ParallelConcatUpdate")
...
 #endif
 
+#ifdef TENSORFLOW_USE_DIRECTML
+
+#define REGISTER_PARALLEL_CONCAT(type)                                     \
+  REGISTER_KERNEL_BUILDER(                                                 \
+      Name("ParallelConcat").Device(DEVICE_DML).TypeConstraint<type>("T"), \
+      FailureKernel);
+REGISTER_PARALLEL_CONCAT(float);
+REGISTER_PARALLEL_CONCAT(Eigen::half);
```
adtsai, line 260: Any reason not to register all the same types CUDA does, since this is a registration-only kernel??

adtsai (approved):
:shipit:

jstoecker (approved):

tensorflow/core/kernels/inplace_ops.cc
PatriceVignola: If we did, this would fail at runtime since _ParallelConcatStart and _ParallelConcatUpdate don't support double. This kernel is never executed, which is why it's a "FailureKernel" that unconditionally fails if constructed. Instead, it's meant to be replaced in one of the graph passes by a mix of _ParallelConcatStart and _ParallelConcatUpdate.

PR #83, opened 2020-11-03
Update BUILD.md to remove reference to unused branch

jstoecker:
Our build doc still references the original two-branch strategy (directml, directml-dev) for DirectML integration with TensorFlow. This change updates the notes to reflect reality, which is that all development occurs in the directml branch and preview APIs may or may not be in use.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -10,7 +10,7 @@ The `directml` branch is considered the main branch for development in this repo
 
 ## Developer Mode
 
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -2,17 +2,29 @@
...
 
 ## Developer Mode
 
-**The python packages produced from the `directml-dev` branch will not run unless you enable developer mode, since they are not intended to be shipped or distributed for mainstream use**. To enable developer mode:
+Changes in this repository may reference in-development or preview versions of DirectML that are subject to change. This allows testing new APIs before they are committed to the DirectML API (once an API appears in DirectML.h it is immutable). A preview build of DirectML requires *developer mode* to be enabled or it will fail to load; this restriction is intended to avoid a long-term dependency on the preview library. **Packages on PyPI will only be released when the repository can depend on a stable version of DirectML.**
```
PatriceVignola: Maybe we should clarify that APIs that appear in the dev versions of the redist are not actually immutable (otherwise, we wouldn't be able to test them before committing them).

PatriceVignola (approved):
:shipit:

PR #84, opened 2020-11-04
register GetSessionHandle, GetSessionHandleV2 and GetSessionTensor for DML

zhangxiang1993:


jstoecker (approved):
Looks good. Just make sure the tests don't implode, since introducing these host-only registrations sometimes persuades TF to start exercising functionality that we haven't implemented. :)

PatriceVignola (approved):

PR #86, opened 2020-11-05
User/xianz/eagerpyfunc

zhangxiang1993:


PatriceVignola (approved):
It looks fine, but again, make sure that it doesn't break tests because of other ops that become enabled.

PR #92, opened 2020-11-17
Merge upstream 1.15.4 changes

jstoecker:
Merges bug/security fixes from upstream 1.15.4. One change to call out is the restriction for numpy in [1.16, 1.19). We've hit a few issues with numpy 1.19.x, so this change moves our test agents over to 1.18.5 as in upstream TF.


third_party/dml/ci/test/run_tests_data.json
```diff
@@ -10,7 +10,7 @@
         "--log_device_placement"
       ],
       "pipDeps": [
-        "numpy==1.19.3",
+        "numpy==1.18.5",
```
PatriceVignola, line 13: Was there a problem with 1.19.3?

jstoecker: No, but there are various parts from the merge that explicitly reference 1.18.5. Setup.py was updated so that the tf pip package depends on version >= 1.6 and < 1.19; rather than revert the upstream change, it seems safer to keep in sync.

PatriceVignola: That's because 1.18.5 is the last version before 1.19.x. However the only break with numpy <1.19.4 was the C API to build TensorFlow, and we cherry-picked the appropriate fix from TensorFlow 2.0.

I don't mind going back to 1.18.5 for our CI machines, but if we do we should probably update the readme warnings to propose 1.18.5 instead of 1.19.3 since we'll have no way to make sure that 1.19.3 isn't broken in other ways that we haven't noticed yet.

jstoecker: Good point on updating the README.md. Users won't even be able to use 1.19.3. It will be incompatible with our package now.

PatriceVignola (approved):

PR #93, opened 2020-11-17
Switch to DirectML 1.4.0

jstoecker:
Use the stable/public DirectML 1.4.0 package. Unfortunately, we'll need to deregister two kernels that require 8D tensor support that isn't in 1.4.0: SpaceToBatch and SpaceToBatchND.

### DIFF
--- a/tensorflow/core/kernels/dml_space_to_batch_op.cc
+++ b/tensorflow/core/kernels/dml_space_to_batch_op.cc
@@ -431,8 +431,9 @@ class DmlSpaceToBatchKernel : public DmlKernel {
       DmlKernelWrapper<DmlSpaceToBatchKernel<SpaceToBatchInitHelper>,   \
                        SpaceToBatchShapeHelper>);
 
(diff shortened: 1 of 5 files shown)


tensorflow/core/kernels/dml_space_to_batch_op.cc
```diff
@@ -431,8 +431,9 @@ class DmlSpaceToBatchKernel : public DmlKernel {
       DmlKernelWrapper<DmlSpaceToBatchKernel<SpaceToBatchInitHelper>,   \
                        SpaceToBatchShapeHelper>);
 
-TF_CALL_float(REGISTER_DML_KERNEL);
-TF_CALL_half(REGISTER_DML_KERNEL);
+// TODO: re-register these kernels when we can depend on DML 1.15.
```
PatriceVignola: Can we add a bug number for this to make sure it doesn't fall through the cracks?

PatriceVignola (approved):

adtsai (approved):
:shipit:

PR #96, opened 2020-11-18
Fix DirectML.dll loading issues

jstoecker:
This change should resolve the DLL loading issues we've encountered with, for example, store versions of python. Previously, we relied on TF augmenting the PATH environment variable in __init__.py. This solution doesn't work in all situations since LoadLibrary only inspects binaries on the PATH for traditional win32/desktop applications. Our WSL packages don't have this problem since they instead modify the RPATH of libtensorflow.so.

This change removes the _solib_directml directory and instead places the DML library in the same folder as the core TF library. The location of the core TF library is different for each platform (within Python site-packages):
- Windows: `tensorflow_core/python/_pywrap_tensorflow_internal.pyd`
- Linux: `tensorflow_core/libtensorflow_framework.so.1`

We continue to use the RPATH for WSL builds, and for Windows we call LoadLibrary with explicit path to the directory containing `_pywrap_tensorflow_internal.pyd`. We could have continued to use a relative path lookup (to _solib_directml or some other directory), but this approach should also work better for builds of the C library. Apps can put both the TF framework and DML redistributable and the same directory and it should simply work.

Finally, this change also removes the extra copies of DML redist files that were being placed in the linux wheels.

### DIFF
--- a/tensorflow/stream_executor/platform/default/dso_loader.cc
+++ b/tensorflow/stream_executor/platform/default/dso_loader.cc
@@ -197,7 +197,13 @@ port::StatusOr<void*> GetDirectMLLibraryHandle(const string& basename) {
   string name = basename;
   if (path.empty()) {
     name += string(".") + DIRECTML_SOURCE_VERSION;
+
+    // Look for DML under the same directory as the core tensorflow module. This
+    // check isn't required for WSL since the RPATH of the tensorflow .so file
+    // is modified.
+#if _WIN32
     path = GetModuleDirectory();
+#endif
   }
 
   return GetDsoHandle(name, "", path);
--- a/tensorflow/stream_executor/platform/default/dso_loader.cc
+++ b/tensorflow/stream_executor/platform/default/dso_loader.cc
@@ -27,6 +27,10 @@ limitations under the License.
 #include "third_party/gpus/cuda/cuda_config.h"
 #include "third_party/tensorrt/tensorrt_config.h"
 
+#if _WIN32
+#include "tensorflow/core/platform/windows/wide_char.h"
+#endif
+
 namespace stream_executor {
 namespace internal {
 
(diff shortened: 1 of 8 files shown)


tensorflow/stream_executor/platform/default/dso_loader.cc
```diff
@@ -41,6 +45,38 @@ string GetDirectMLPath() {
...
+      reinterpret_cast<LPCWSTR>(&GetModuleDirectory), &tensorflowHmodule);
+  CHECK_EQ(getHandleResult, TRUE);
+
+  // Safe const_cast because of explicit bounds checking and contiguous memory
+  // in C++11 and later.
+  std::wstring wpath(MAX_PATH, '\0');
+  DWORD filePathSize =
+      GetModuleFileNameW(tensorflowHmodule, const_cast<wchar_t*>(wpath.data()),
+                         static_cast<DWORD>(wpath.size()));
+  if (filePathSize > wpath.size()) {
```
adtsai: >filePathSize > wpath.size() [](start = 6, length = 27)

This should check if GetLastError() == ERROR_INSUFFICIENT_BUFFER and use a loop, doubling the buffer each iteration. This is because if the buffer is too small, GetModuleFileName(h, p, len) returns len.

tensorflow/stream_executor/platform/default/dso_loader.cc
```diff
@@ -41,6 +45,38 @@ string GetDirectMLPath() {
...
+    wpath.resize(filePathSize);
+    filePathSize = GetModuleFileNameW(tensorflowHmodule,
+                                      const_cast<wchar_t*>(wpath.data()),
+                                      static_cast<DWORD>(wpath.size()));
+  }
+  CHECK_NE(filePathSize, 0);
+
+  // Strip TF library filename from the path.
+  size_t pathDividerOffset = wpath.find_last_of('\\');
```
adtsai: [https://docs.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchremovefilespec](https://docs.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchremovefilespec)

tensorflow/stream_executor/platform/default/dso_loader.cc
jstoecker: Nice catch! You found _two_ bugs, since this isn't original code. :)

tensorflow/stream_executor/platform/default/dso_loader.cc
```diff
@@ -41,6 +46,42 @@ string GetDirectMLPath() {
...
+    filePathSize = GetModuleFileNameW(tensorflowHmodule,
+                                      const_cast<wchar_t*>(wpath.data()),
+                                      static_cast<DWORD>(wpath.size()));
+  }
+  CHECK_NE(filePathSize, 0);
+
+  // Strip TF library filename from the path.
+  CHECK_EQ(
+      PathCchRemoveFileSpec(const_cast<wchar_t*>(wpath.data()), wpath.size()),
+      S_OK);
```
adtsai, line 81: >S_OK [](start = 6, length = 4)

You should use SUCCEEDED, as this function could return S_FALSE.

tensorflow/stream_executor/platform/default/dso_loader.cc
```diff
@@ -41,6 +46,42 @@ string GetDirectMLPath() {
...
+    wpath.resize(wpath.size() * 2);
+    filePathSize = GetModuleFileNameW(tensorflowHmodule,
+                                      const_cast<wchar_t*>(wpath.data()),
+                                      static_cast<DWORD>(wpath.size()));
+  }
+  CHECK_NE(filePathSize, 0);
+
+  // Strip TF library filename from the path.
+  CHECK_EQ(
+      PathCchRemoveFileSpec(const_cast<wchar_t*>(wpath.data()), wpath.size()),
```
adtsai, line 80: >PathCchRemoveFileSpec [](start = 6, length = 21)

You should explicitly truncate the string afterwards with `wpath.resize(wcslen(wpath.c_str()))`

tensorflow/stream_executor/platform/default/dso_loader.cc
jstoecker: This is intentional. S_FALSE indicates there was nothing to remove, which should not happen if it's a valid path to a module.

adtsai (approved):
:shipit:

PR #97, opened 2020-11-18
Stop gathering device placement logs during CI

PatriceVignola:
We don't use this data anymore, so we can save quite a bit of time during CI by disabling device placement log gathering.


jstoecker (approved):

jstoecker:
Can you cherry pick this into release/2020-11 as well?

PR #102, opened 2020-11-18
Use compute queue for AMD devices

PatriceVignola:
All the evidence we've seen so far points to compute queues being better at preventing TDRs, but also being more performant on AMD. We can always revert the change later if it turns out to not be stable enough, but we should at least have this change be tested by Autopilot and ai-benchmark tests.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_device_state.cc
+++ b/tensorflow/core/common_runtime/dml/dml_device_state.cc
@@ -70,8 +70,14 @@ namespace tensorflow {
   ComPtr<IDMLDevice> dml_device;
   dml_device = CreateDmlDevice(d3d_device.Get(), dml_flags);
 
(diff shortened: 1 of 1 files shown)


jstoecker (approved):

tensorflow/core/common_runtime/dml/dml_device_state.cc
```diff
@@ -70,8 +70,14 @@ namespace tensorflow {
   ComPtr<IDMLDevice> dml_device;
   dml_device = CreateDmlDevice(d3d_device.Get(), dml_flags);
 
+  D3D12_COMMAND_LIST_TYPE queue_type = D3D12_COMMAND_LIST_TYPE_DIRECT;
+
+  if (adapter.VendorID() == VendorID::kAmd) {
+    queue_type = D3D12_COMMAND_LIST_TYPE_COMPUTE;
```
jstoecker: Let's keep an environment variable to force the queue on/off as well, but default to on for AMD. Might be useful for experimentation.

PR #109, opened 2020-12-03
Fix spurious OOMs on integrated GPUs

adtsai:
Fixes reported OOMs on integrated GPUs when per_process_gpu_memory_fraction is set. The cause of this bug was that we were only considering dedicated memory when computing the total GPU memory available, and integrated GPUs don't have any dedicated memory. Now, for UMA adapters we will include shared memory as part of a GPU's total memory.


jstoecker (approved):
🥇

jstoecker:
We should consider pushing this to release/2020-11 as well.

PR #114, opened 2020-12-07
Fix list kernels and tensor array ops registration

PatriceVignola:
Tensor array template helpers had default implementations because Variant and Resource don't have eigen implementations. The problem with having default implementations though is that the compiler can't tell us if we forgot to add registrations for new datatypes, which caused regression for DML, CPU and GPU devices when refactoring the List kernels recently.

The new approach that we're taking is to remove the default implementation and to instead explicitly register the unsupported types (e.g. Variant and Resource) to the old default implementation. This will make sure that the compiler will stop us in the future when we add kernels that use different datatypes.

This change also fixes lingering GPU issues since we haven't built the GPU flavor in a while.

### DIFF
--- a/build.py
+++ b/build.py
@@ -200,7 +200,7 @@ def build(args):
   if args.subcommands:
     cl.append("--subcommands")
   cl.append("--config=opt")
-  cl.append("--config=dml")
+  cl.append("--config=cuda")
   if args.telemetry:
     cl.append("--config=dml_telemetry")
   if args.config == "debug":
--- a/build.py
+++ b/build.py
@@ -200,7 +200,7 @@ def build(args):
   if args.subcommands:
     cl.append("--subcommands")
   cl.append("--config=opt")
-  cl.append("--config=cuda")
+  cl.append("--config=dml")
   if args.telemetry:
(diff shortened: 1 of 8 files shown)


jstoecker:
It's possible to delete default template implementations in C++11 (e.g. [`template <typename T> GUID uuidof() = delete;`](https://github.com/microsoft/tensorflow-directml/blob/8d08cf53602de5372b4a20eb069ff347e3d0ce86/third_party/dml/winadapter/dxguids/dxguids.h#L28)). This would produce a compiler error if we don't specialize for certain types. Any reason we didn't do that?

PatriceVignola:
> It's possible to delete default template implementations in C++11 (e.g. [`template <typename T> GUID uuidof() = delete;`](https://github.com/microsoft/tensorflow-directml/blob/8d08cf53602de5372b4a20eb069ff347e3d0ce86/third_party/dml/winadapter/dxguids/dxguids.h#L28)). This would produce a compiler error if we don't specialize for certain types. Any reason we didn't do that?

Good point. Better to fail at compile time than linking time :) Fixed!

jstoecker (approved):

PR #118, opened 2020-12-08
Pipeline should continue if a test job fails

jstoecker:
Test jobs that fail because of test agents crashing, timing out, or otherwise failing shouldn't mark the entire build as failed.


adtsai (approved):
:shipit:

PR #121, opened 2020-12-12
Implement all the arithmetic Scatter and ResourceScatter operators

PatriceVignola:
The arithmetic Scatter and ResourceScatter operators are used quite a lot in the TensorFlow python tests, so they create a lot of false positive failures. Implementing these ups gets rid of about 50% of the failures and should improve performance on some models that we track.

### DIFF
--- a/tensorflow/core/kernels/dml_ops_common.cc
+++ b/tensorflow/core/kernels/dml_ops_common.cc
@@ -351,7 +351,7 @@ DML_SCALAR_UNION ScalarTensor(double value, DML_TENSOR_DATA_TYPE data_type) {
       break;
 
     default:
-      DML_CHECK_SUCCEEDED(E_NOTIMPL);
+      DML_CHECK_SUCCEEDED(E_INVALIDARG);
       break;
   }
 
--- a/tensorflow/core/kernels/dml_ops_common.cc
+++ b/tensorflow/core/kernels/dml_ops_common.cc
@@ -304,3 +304,57 @@ const DML_BUFFER_BINDING* DmlKernel::GetPersistentResourceBinding() const {
 }
 
 }  // namespace tensorflow
+
+namespace dml {
+DML_SCALAR_UNION ScalarTensor(double value, DML_TENSOR_DATA_TYPE data_type) {
+  DML_SCALAR_UNION scalar;
+
+  switch (data_type) {
+    case DML_TENSOR_DATA_TYPE_INT8:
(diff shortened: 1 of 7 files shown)


tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -304,3 +304,57 @@ const DML_BUFFER_BINDING* DmlKernel::GetPersistentResourceBinding() const {
 }
 
 }  // namespace tensorflow
+
+namespace dml {
+DML_SCALAR_UNION ScalarTensor(double value, DML_TENSOR_DATA_TYPE data_type) {
```
jstoecker, line 309: nit: rename helper function to `ScalarUnion` since it doesn't return a tensor

jstoecker: nit: rename helper function to `ScalarUnion` since this doesn't return a tensor

tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -304,3 +304,57 @@ const DML_BUFFER_BINDING* DmlKernel::GetPersistentResourceBinding() const {
...
+
+    case DML_TENSOR_DATA_TYPE_INT64:
+      scalar.Int64 = static_cast<int64_t>(value);
+      break;
+
+    case DML_TENSOR_DATA_TYPE_UINT64:
+      scalar.UInt64 = static_cast<uint64_t>(value);
+      break;
+
+    case DML_TENSOR_DATA_TYPE_FLOAT32:
```
jstoecker, line 345: Add DML_TENSOR_DATA_TYPE_FLOAT16?

third_party/dml/DirectMLX.h
```diff
@@ -2512,9 +2512,31 @@ namespace dml
     // TODO: GatherND (DML_OPERATOR_GATHER_ND1)
     // 
 
-    // 
-    // TODO: ScatterElements (DML_SCATTER_ELEMENTS_OPERATOR_DESC)
-    // 
+    inline Expression ScatterElements(
```
jstoecker, line 2515: This reminds me that we should transition to the copy of [DirectMLX.h in the DirectML github repo](https://github.com/microsoft/DirectML/blob/master/Libraries/DirectMLX.h). I created bug 31071585 for this. In the meantime, make sure to port this change over to GitHub.

jstoecker (approved):

PR #125, opened 2020-12-15
Register SqrtGrad and RsqrtGrad kernels

adtsai:
SqrtGrad is used by the NVIDIA-SPADE model.

### DIFF
--- a/tensorflow/core/kernels/dml_cwise_ops.cc
+++ b/tensorflow/core/kernels/dml_cwise_ops.cc
@@ -707,6 +707,10 @@ REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(SigmoidGrad, y* x*(1 - x),
                                            kNchwDimensionCount)
 REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(TanhGrad, y*(1 - x * x),
                                            kNchwDimensionCount)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(SqrtGrad, y * 0.5f / x,
+                                           kNchwDimensionCount)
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/dml_cwise_ops.cc
```diff
@@ -707,6 +707,10 @@ REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(SigmoidGrad, y* x*(1 - x),
                                            kNchwDimensionCount)
 REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(TanhGrad, y*(1 - x * x),
                                            kNchwDimensionCount)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(SqrtGrad, y * 0.5f / x,
+                                           kNchwDimensionCount)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(RsqrtGrad, y * (-0.5f * x) * (x * x),
```
PatriceVignola, line 712: Do we know if identity(x) * x * x faster than identity(x^3)?

jstoecker (approved):

tensorflow/core/kernels/dml_cwise_ops.cc
jstoecker: DML pow with a constant uses HLSL pow intrinsic; pow with an integer exponent tensor uses repeated squaring and would be the same as this (fewer dispatches though). I'm not really sure which would be faster in this case...

adtsai: In lieu of evidence for either case, I'm inclined to leave it as-is. :)

PR #127, opened 2020-12-17
Implement unsorted segment reduction ops for DML devices

PatriceVignola:


tensorflow/core/kernels/dml_segment_reduction_ops.cc
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  if (reduce_function == DML_REDUCE_FUNCTION_MIN) {
+    return std::numeric_limits<TData>::max();
+  }
+
+  if (reduce_function == DML_REDUCE_FUNCTION_MAX) {
+    return std::numeric_limits<TData>::lowest();
+  }
+}
```
adtsai, line 96: Might want to assert or abort or something in case somebody accidentally supplies a reduce function that isn't one of these four.

tensorflow/core/kernels/dml_segment_reduction_ops.cc
```diff
@@ -0,0 +1,255 @@
...
+  REGISTER_UNSORTED_SEGMENT_REDUCTION_KERNEL(type, "UnsortedSegmentMin",   \
+                                             SegmentReductionMinOp<type>); \
+  REGISTER_UNSORTED_SEGMENT_REDUCTION_KERNEL(type, "UnsortedSegmentProd",  \
+                                             SegmentReductionProdOp<type>);
+
+TF_CALL_float(REGISTER_UNSORTED_SEGMENT_REDUCTION_DML_KERNEL);
+TF_CALL_half(REGISTER_UNSORTED_SEGMENT_REDUCTION_DML_KERNEL);
+TF_CALL_int32(REGISTER_UNSORTED_SEGMENT_REDUCTION_DML_KERNEL);
+#undef REGISTER_DML_KERNEL
+#undef REGISTER_UNSORTED_SEGMENT_REDUCTION_KERNEL
```
adtsai, line 251: #undef REGISTER_UNSORTED_SEGMENT_REDUCTION_DML_KERNEL?

adtsai (approved):
:shipit:

tensorflow/core/kernels/dml_segment_reduction_ops.cc
PatriceVignola: It won't compile then because it's a constexpr function and needs to return something at compile-time.

jstoecker: It's not as easy to read, but you can use template specialization (with classes not functions):

```cpp
template <DML_REDUCE_FUNCTION reduce_function, typename TData> 
struct IdentityValueFunctor { constexpr TData operator()() = delete; };

template <typename TData>
struct IdentityValueFunctor<DML_REDUCE_FUNCTION_MULTIPLY, TData>
{ constexpr TData operator()() { return static_cast<TData>(1); } };

template <typename TData>
struct IdentityValueFunctor<DML_REDUCE_FUNCTION_SUM, TData>
{ constexpr TData operator()() { return static_cast<TData>(0); } };

template <typename TData>
struct IdentityValueFunctor<DML_REDUCE_FUNCTION_MIN, TData>
{ constexpr TData operator()() { return std::numeric_limits<TData>::max(); } };

template <typename TData>
struct IdentityValueFunctor<DML_REDUCE_FUNCTION_MAX, TData>
{ constexpr TData operator()() { return std::numeric_limits<TData>::lowest(); } };
```

jstoecker (approved):

tensorflow/core/kernels/dml_segment_reduction_ops.cc
PatriceVignola: Ah, what I meant is that we don't need to assert or abort because it won't even compile if something calls this function with a different reduce function. So I went for the more compact and easier to read approach instead of template specializations.

PR #128, opened 2020-12-17
Decouple Select and SelectV2's validation

PatriceVignola:
Select and SelectV2 have different validation and broadcasting rules; therefore, they shouldn't use the same initialization helper.


adtsai (approved):
:shipit:

jstoecker (approved):

PR #133, opened 2020-12-18
Move binding logic outside of execution context lock

adtsai:
Recall that although our kernels execute in a free-threaded manner, at some point they need to dispatch to the device. Currently, all of the dispatching logic occurs under the execution context lock. This is because dispatches need to be serialized to ensure correct sequencing of work.

However, not all of this work needs to be under the lock. It's true that the *dispatches* need to be sequenced, but the *binding* of resources to those dispatches does not. Binding is a relatively expensive operation, with most of the CPU time spent inside the UMD creating descriptors.

The DML API is explicitly designed to allow free-threaded binding of operators (and even free-threaded recording of dispatches, though that requires multiple command lists and is a discussion for another time.) This change moves the work of binding input/output/temporary/persistent resources out of the lock, so that the binding work of each operator can run in parallel. This can result in substantial gains in performance in CPU-bound scenarios.

After this change, we expect a ~20-25% reduction in training time for models like LSTM-Sentiment and Pixel-RNN.

* Most binding work (for execution) has been moved out of the command recorder and into the kernel itself.
* Descriptors are now allocated using a BFC allocator rather than a ring-buffer DescriptorPool. This is because allocations and deallocations are no longer synchronized underneath a lock, so allocation of descriptors needs to be able to handle allocates and frees that happen out of order.
* The event queue has been augmented to process callbacks in batches (an optimization) and also to handle callbacks which arrive out-of-order with respect to their desired fence value (necessary to prevent deadlocks). Like the change to descriptor allocation, this latter point is necessary because the enqueueing of these events are no longer serialized and can now arrive in any order.
* Fixes a couple of possible race conditions in dml_where_op and dml_check_numerics_op.
* Fixes a latent bug during operator initialization. Strictly speaking, the operator initializer shouldn't be released or reset until all executions using that initializer have completed execution on the GPU. This has been fixed by allocating a new initializer for each operator initialization.


tensorflow/core/common_runtime/dml/dml_command_recorder.h
```diff
@@ -26,21 +26,29 @@ namespace tensorflow {
...
 
 class DmlCommandRecorder {
  public:
   DmlCommandRecorder(ID3D12Device* d3d_device, IDMLDevice* device,
                      std::shared_ptr<DmlCommandQueue> command_queue,
                      DmlAllocator* allocator);
 
+  // Unlike ExecuteOperator(), this will set up the descriptors and binding
+  // table for you, and automatically release them (using the given event queue)
+  // whenever the initialization completes on the GPU.
```
jstoecker: This asymmetry with `ExecuteOperator` is a little odd. Is there a benefit to having the `IDMLOperatorInitializer` created and destroyed by the callback enqueued inside this method instead of a callback enqueued by `DmlKernel::Initialize` (i.e. the same way you manage the lifetime of the binding table for execution)?

jstoecker (approved):

tensorflow/core/common_runtime/dml/dml_command_recorder.h
adtsai: Hmm, yes, that's a good point - not sure why I didn't think of that.

PR #137, opened 2020-12-29
Register DML's QR kernel for CPU memory

PatriceVignola:
If the cuSolver bug ever get fixed we may want to implement our own DML kernel to stay on par with CUDA, but in the meantime it's probably not worth it to spend time on this kernel. Registering this kernel with CPU memory for DML devices fixes co-location issues seen with some Random kernels (e.g. RandomStandardNormal).

### DIFF
--- a/tensorflow/core/kernels/qr_op_float.cc
+++ b/tensorflow/core/kernels/qr_op_float.cc
@@ -33,4 +33,20 @@ REGISTER_KERNEL_BUILDER(Name("Qr")
                         QrOp<float>);
 #endif
 
+#ifdef TENSORFLOW_USE_DIRECTML
+// For now, register QR as a CPU kernel for DML to follow the same registration
+// as the GPU. If the cuSolver bug ever get fixed we may want to implement our
+// own DML kernel to stay on par with CUDA, but in the meantime it's probably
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/qr_op_float.cc
```diff
@@ -33,4 +33,20 @@ REGISTER_KERNEL_BUILDER(Name("Qr")
                         QrOp<float>);
 #endif
 
+#ifdef TENSORFLOW_USE_DIRECTML
+// For now, register QR as a CPU kernel for DML to follow the same registration
+// as the GPU. If the cuSolver bug ever get fixed we may want to implement our
+// own DML kernel to stay on par with CUDA, but in the meantime it's probably
+// not worth it to spend time on this kernel. Registering this kernel with CPU
```
jstoecker, line 40: I'm not sure I follow why we're registering this for CPU. It's a slippery slope to register ops we are unlikely to implement to influence placement.

Generator ops (i.e. ops with no inputs that only produce outputs), like RandomStandardNormal, are placed on the same device as their consumers (assuming all consumers are on the same device). Are you trying to force the random ops to run on DML in some model?

jstoecker (changes requested):

PR #138, opened 2020-12-29
Link disabled test to grappler optimization task

PatriceVignola:
DML devices don't have auto-mixed precision fast paths in grappler yet, so the float32 -> float16 overflow test doesn't work for us since it would never overflow.

### DIFF
--- a/tensorflow/python/training/experimental/mixed_precision_test.py
+++ b/tensorflow/python/training/experimental/mixed_precision_test.py
@@ -117,7 +117,10 @@ class MixedPrecisionTest(test.TestCase, parameterized.TestCase):
     self.assertFalse(config.get_optimizer_experimental_options()
                      .get('auto_mixed_precision', False))
 
-  # TFDML #25510022
+  # DML doesn't have float16 optimizations in grappler yet, so we wouldn't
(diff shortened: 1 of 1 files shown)


tensorflow/python/training/experimental/mixed_precision_test.py
```diff
@@ -117,7 +117,10 @@ def test_optimizer_errors(self):
     self.assertFalse(config.get_optimizer_experimental_options()
                      .get('auto_mixed_precision', False))
 
-  # TFDML #25510022
+  # DML doesn't have float16 optimizations in grappler yet, so we wouldn't
+  # overflow in this case
+  # TODO #26542950: Enable Tensorcore specific grappler optimizations
+  # TFDML #26542950
```
jstoecker, line 123: Was this change supposed to disable something new, or just add comments?

PatriceVignola: Just adding comments (I should rename the PR). Now that the cause for the disabled test is known, we can close the old generic bug that was created to investigate the failure and instead link it to the grappler optimization task.

jstoecker (approved):

PR #139, opened 2020-12-29
Fix shape simplification for Pad

PatriceVignola:
We use more aggressive shape simplification than the CPU for Pad, but this can only really work for Constant padding. For Reflect and Symmetric padding, we cannot collapse padded dimensions with non-padded dimensions since the padding values depend on the shape.

### DIFF
--- a/tensorflow/core/kernels/dml_pad_op.cc
+++ b/tensorflow/core/kernels/dml_pad_op.cc
@@ -49,6 +49,7 @@ struct SimplePad {
 template <typename Tpadding>
 absl::optional<SimplePad> SimplifyPad(const TensorShape& input_shape,
                                       const Tensor& paddings_tensor,
+                                      bool constant_pad,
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/dml_pad_op.cc
```diff
@@ -49,6 +49,7 @@ struct SimplePad {
 template <typename Tpadding>
 absl::optional<SimplePad> SimplifyPad(const TensorShape& input_shape,
                                       const Tensor& paddings_tensor,
+                                      bool constant_pad,
                                       size_t min_output_size = 4,
                                       size_t max_output_size = 5) {
```
jstoecker: Can we increase this to 8 now?

adtsai: Especially since the error message was updated to 8D, but this constant hasn't.

adtsai (approved):

PR #143, opened 2021-01-02
Fix some device placement tests

PatriceVignola:


tensorflow/python/eager/device_placement_test.py
```diff
@@ -45,12 +43,10 @@ def testDefaultPlacement(self):
     self.assertIn(test_util.gpu_device_type(), c.device)
     self.assertIn('CPU', d.device)
 
-  # TFDML #25509663
-  @test_util.skip_dml
   @test_util.run_gpu_only
   def testUnsupportedDevice(self):
     gpu_type = test_util.gpu_device_type()
-    a = constant_op.constant(1.0)
+    a = constant_op.constant(1)
```
jstoecker, line 49: Looks like this was fixed upstream as well. It might be worth doing a diff of the python tests in 1.15 against upstream master in case there are other fixes like this; however, it might also be a little painful with our edits and potentially new changes that don't work in 1.15. :)

jstoecker (approved):

adtsai (approved):

tensorflow/python/eager/device_placement_test.py
PatriceVignola: This was never a problem in upstream to begin with, this is a bug that I introduced a while ago when onboarding the tests (don't ask me how/why I changed 1 for 1.0, I have no idea :D)

jstoecker: 🤦

PR #145, opened 2021-01-05
Remove potential aborts from DML device enumeration

jstoecker:
DML devices are enumerated at startup, so we should be very careful about aborting with a fatal error in this stage: it prevents the use of CPU devices (e.g. consider Lobe). In particular, TFDML will fail in `CreateDXGIFactory` when running on Windows 10 version 1709 because `IDXGIFactory6` wasn't introduced until version 1803. This change will try to enumerate adapters based on GPU preference, but it will fall back to using unordered adapters if necessary. And, if all else fails, it will return an empty list of adapters.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_adapter_impl.cc
+++ b/tensorflow/core/common_runtime/dml/dml_adapter_impl.cc
@@ -169,25 +169,20 @@ uint64_t DmlAdapterImpl::QueryAvailableLocalMemory() const {
 }
 
 std::vector<DmlAdapterImpl> EnumerateAdapterImpls() {
-  ComPtr<IDXGIFactory6> dxgi_factory;
-  DML_CHECK_SUCCEEDED(CreateDXGIFactory(IID_PPV_ARGS(&dxgi_factory)));
+  ComPtr<IDXGIFactory4> dxgi_factory;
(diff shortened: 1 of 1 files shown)


PatriceVignola (approved):

tensorflow/core/common_runtime/dml/dml_adapter_impl.cc
```diff
@@ -169,46 +169,72 @@ uint64_t DmlAdapterImpl::QueryAvailableLocalMemory() const {
...
 
-  const D3D_FEATURE_LEVEL min_feature_level = D3D_FEATURE_LEVEL_11_0;
   std::vector<DmlAdapterImpl> adapter_infos;
 
-  uint32_t adapter_index = 0;
-  ComPtr<IDXGIAdapter1> adapter;
-  while (dxgi_factory->EnumAdapterByGpuPreference(
-             adapter_index, DXGI_GPU_PREFERENCE_HIGH_PERFORMANCE,
-             IID_PPV_ARGS(&adapter)) != DXGI_ERROR_NOT_FOUND) {
+  auto IsSoftwareAdapter = [](IDXGIAdapter1* adapter) -> bool {
```
adtsai: This seems like it'd be better off just being its own function, rather than being stuffed into EnumerateAdapterImpls.

adtsai (approved):

PR #148, opened 2021-01-06
Fix GetLocalDMLInfo calls in grappler optimizers

jstoecker:
Some parts of grappler look up physical device properties when clustering or building a cost model. Our current implementation assumes that TF device IDs map directly to DX adapter IDs, which is _usually_ correct but not always. It's easy to break the grappler simply by using virtual device mappings (i.e. treating a single physical device as multiple logical devices), and reordering the visible device list will result in the wrong hardware properties being used.

Sadly, the solution to this problem is imperfect: the grappler doesn't have enough context to correctly map device IDs to adapter IDs in all cases because visible device lists are unique to each TF session. We adopt the same (unfortunate) solution that GPU devices use: maintain device ID to adapter ID mappings in a process-wide singleton and raise an error if a client tries to reorder visible device mappings.


adtsai (approved):
:shipit:

PR #151, opened 2021-01-07
Register ZerosLike with Variant datatype

PatriceVignola:
The GPU supports 2 types of variants: OptionalVariant and TensorList. Therefore, we should support the same variants for DML. This fixes a few test failures caused by not having variant registrations for ZerosLike.


adtsai (approved):
:shipit:

jstoecker (approved):

PR #152, opened 2021-01-07
Fix reduce special cases when the input is empty but the output is non-empty

PatriceVignola:
We were assuming that the output would always be a scalar in this case, but it can be a tensor of any dimension.

### DIFF
--- a/tensorflow/core/kernels/dml_reduce_ops.cc
+++ b/tensorflow/core/kernels/dml_reduce_ops.cc
@@ -175,11 +175,10 @@ class DmlReduceKernel : public DmlKernel {
       DmlKernelTensors tensors;
       tensors.outputs.resize(1);
 
-      constexpr uint32_t dml_output_shape[] = {1u};
-
       tensors.outputs[0].emplace();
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/dml_reduce_ops.cc
```diff
@@ -175,11 +175,10 @@ class DmlReduceKernel : public DmlKernel {
       DmlKernelTensors tensors;
       tensors.outputs.resize(1);
 
-      constexpr uint32_t dml_output_shape[] = {1u};
-
       tensors.outputs[0].emplace();
-      tensors.outputs[0]->desc = DmlTensorDesc::Create(
-          output_type, dml_output_shape, dml_output_shape);
+      tensors.outputs[0]->desc =
+          DmlTensorDesc::Create(output_type, ctx->GetOutputTensorShape(0),
```
jstoecker: I assume the output shape might be something like `[1,1,1,1,1,1,1,1,1,1,1]`? It seems like we may want some shape simplification for the purpose of DML's tensor desc to ensure we don't run into a situation where the op can't be created.

jstoecker (approved):

PR #155, opened 2021-01-08
Untangle DML/GPU in remapper and add _FusedMatMul for DML

jstoecker:
The grappler remapper should treat GPU and DML devices differently, so this change splits out logic that was previously nested into the existing GPU functions into separate DML functions. I kept the existing restrictions on conv2d fusions that apply to GPU devices, but we should investigate if these are relevant for DML devices as well (opened bug 31294633).

This change also enables _FusedMatMul (MatMul + Bias + [Activation]), which is  a straight mapping to DML's GEMM operator, and allows float16 fusions for the fused conv kernel.

### DIFF
--- a/tensorflow/core/kernels/dml_conv_ops.cc
+++ b/tensorflow/core/kernels/dml_conv_ops.cc
@@ -659,7 +659,8 @@ class DmlFusedConv2DKernel : public DmlKernel {
   REGISTER_KERNEL_BUILDER(                                               \
       Name("_FusedConv2D").Device(DEVICE_DML).TypeConstraint<type>("T"), \
       DmlKernelWrapper<DmlFusedConv2DKernel<type>, ConvShapeHelper>);
-TF_CALL_DML_FLOAT_TYPES(DML_REGISTER_KERNEL);
+// FusedConv2D only supports float32
+TF_CALL_float(DML_REGISTER_KERNEL);
 #undef DML_REGISTER_KERNEL
 
 class DmlConv2DBackpropInputKernel : public DmlKernel {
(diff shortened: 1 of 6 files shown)


adtsai:
> 

Does this change inadvertently disable the fusion to FusedBatchNormEx?

---
Refers to: tensorflow/core/grappler/optimizers/remapper.cc:1035 in d80e6b3. [](commit_id = d80e6b342304341dddc29b2e38ea5c8a4a6baf06, deletion_comment = False)

tensorflow/core/kernels/dml_conv_ops.cc
```diff
@@ -659,8 +659,7 @@ class DmlFusedConv2DKernel : public DmlKernel {
   REGISTER_KERNEL_BUILDER(                                               \
       Name("_FusedConv2D").Device(DEVICE_DML).TypeConstraint<type>("T"), \
       DmlKernelWrapper<DmlFusedConv2DKernel<type>, ConvShapeHelper>);
-// FusedConv2D only supports float32
-TF_CALL_float(DML_REGISTER_KERNEL);
+TF_CALL_DML_FLOAT_TYPES(DML_REGISTER_KERNEL);
 #undef DML_REGISTER_KERNEL
```
adtsai: Does this work? The opdef indicates that this only supports float and double.

adtsai: (similar comment applies to _FusedMatMul)

jstoecker: Ah you're right, I didn't check the opdef. Strange restriction.... I'll revert to float only.

jstoecker:
> > 
> 
> Does this change inadvertently disable the fusion to FusedBatchNormEx?
> 
> Refers to: tensorflow/core/grappler/optimizers/remapper.cc:1035 in [d80e6b3](https://github.com/microsoft/tensorflow-directml/commit/d80e6b342304341dddc29b2e38ea5c8a4a6baf06). [](commit_id = [d80e6b3](https://github.com/microsoft/tensorflow-directml/commit/d80e6b342304341dddc29b2e38ea5c8a4a6baf06), deletion_comment = False)

Good catch. This is why it's evil to treat "DML" and "GPU" as the same.

adtsai (approved):

PR #156, opened 2021-01-13
Implement depthwise conv grad ops

jstoecker:


jstoecker:
Noticed a few bugs. Will reopen when I address them.

PR #158, opened 2021-01-14
Fix Grappler cost model graph fetching for DML

PatriceVignola:
We were appending a JSON-encoded string at the end of the device name that we insert into the NodeExecStatsInterface, but then grappler was failing many comparisons since DML:0 != DML:0 ({json_data...})

### DIFF
--- a/tensorflow/core/common_runtime/executor.cc
+++ b/tensorflow/core/common_runtime/executor.cc
@@ -2215,20 +2215,7 @@ bool ExecutorState::NodeDone(const Status& s, const Node* node,
   nodestats::SetAllEnd(stats);
   if (stats) {
     if (stats_collector_) {
-#if TENSORFLOW_USE_DIRECTML
-      const auto& physical_device_desc =
-          impl_->params_.device->attributes().physical_device_desc();
-      if (!physical_device_desc.empty()) {
-        auto device_name = absl::StrCat(impl_->params_.device->name(), " (",
-                                        physical_device_desc, ")");
-        stats->Done(device_name);
-      }
-      else {
-        stats->Done(impl_->params_.device->name());
-      }
-#else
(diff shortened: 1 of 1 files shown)


tensorflow/core/common_runtime/executor.cc
```diff
@@ -2215,20 +2215,7 @@ bool ExecutorState::NodeDone(const Status& s, const Node* node,
   nodestats::SetAllEnd(stats);
   if (stats) {
     if (stats_collector_) {
-#if TENSORFLOW_USE_DIRECTML
```
PatriceVignola, line 2218: @adtsai Is there a reason this change was needed in the first place? I couldn't find where this is being parsed or being used for DML-specific stuff. If it's being used somewhere, it would likely have stopped working anyway when you changed physical_device_desc to be json data instead of a simple string.

adtsai: This was done for the benefit of chrome traces. After the json change the string looks a little different, but it still shows up in the trace:

image

What gets broken by this change?

PatriceVignola: I don't think we're meant to modify this string at all since it's not only visual but is used logically in the code. This "name" is used as the key in a dictionary in the cost model graph logic failing comparisons when compared against the device strings in other dictionaries (e.g. the vanilla /device/DML:0 strings). See StepStatsCollector::FinalizeInternal in step_stats_collector.cc. FunctionTest.testSwapImplementationWithGrapplerPlugin was failing because of this.

adtsai: Hmm, interesting. I didn't know this was used for any sort of logic. I guess we'll have to go without prettified device names in the chrome traces.

adtsai (approved):
:shipit:

PR #159, opened 2021-01-14
Fix float->bool DML cast

PatriceVignola:


adtsai (changes requested):
What does `ceil(-0.5f)` return?

PatriceVignola:
> What does `ceil(-0.5f)` return?

Good one :D

adtsai (approved):
:shipit:

PR #161, opened 2021-01-19
Implement depthwise conv grad ops

jstoecker:
Implements the backpropagation operators for the input and filter of depthwise 2D convolution:
- DepthwiseConv2dNativeBackpropFilter
- DepthwiseConv2dNativeBackpropInput

Note that the depthwise filter backprop kernel is a little different from the standard 2D version: to express the math as a convolution we actually need to use a 3D convolution. There are at least two alternative implementations I can think of, but they have drawbacks:
1. Expressing the calculation as a series of 2D convolutions with their outputs joined.
2. Transposing dimensions physically (i.e. copying with identity op and strides) and then coalescing some of them for a 2D convolution.

There various 2D conv kernels share quite a bit in common and could be refactored in the future, but I am not considering that for this change. Most of the code is validation logic anyway, which we want to keep similar to existing TF kernels.


adtsai (approved):
:shipit:

PR #162, opened 2021-01-19
User/xianz/scatter nd

zhangxiang1993:


tensorflow/core/kernels/dml_resource_scatter_ops.cc
```diff
@@ -0,0 +1,437 @@
...
+
+    auto status_or_event =
+        DmlKernel::Compute(ctx, input_bindings, output_bindings);
+    if (!status_or_event.ok()) {
+      return status_or_event;
+    }
+
+    gpu_event = status_or_event.ValueOrDie();
+
+    init_helper->Unlock();
```
jstoecker: This is super sketchy: there are some returns earlier in the function that would prevent the mutex from unlocking. One could argue that an error here would be fatal anyway, but I'd feel better if we had an RAII type that unlocks when destroyed (i.e. regardless of how compute returns).

It looks like this is an issue in the dml_scatter_update_ops.cc file as well (presumably adopted here), so it would be great to see both places updated. It does look like these kernels are never cached, which is good, so we shouldn't see multiple unlocks at least.

jstoecker (changes requested):
Looks good aside from one issue!

jstoecker (approved):

PR #163, opened 2021-01-20
Remove potential aborts from DML device enumeration (#145)

jstoecker:
Cherry pick isolated bug fix into release branch for Lobe.


adtsai (approved):
If this is going into the official release/2020-11 branch, we should presumably also produce a new package and put it up on PyPi (it'd be a little weird otherwise to have changes sitting in a release branch that don't ever go out into a release).

jstoecker:
> If this is going into the official release/2020-11 branch, we should presumably also produce a new package and put it up on PyPi (it'd be a little weird otherwise to have changes sitting in a release branch that don't ever go out into a release).

Maybe... we won't really service this long term since it's all pre-release anyway, so I'm not too concerned. Let's see if there are any other changes needed between now and the next release out of master before publishing something to pypi.

PR #167, opened 2021-01-25
Implement 3D conv ops

jstoecker:
Adds kernels for the following ops:
- Conv3D
- Conv3DBackpropInput
- Conv3DBackpropInputV2
- Conv3DBackpropFilter
- Conv3DBackpropFilterV2

There are two differences between the "V1" and V2 gradient ops:
1. The V2 ops receive the non-backprop tensor as a shape tensor rather than the original tensor from the forward pass. For example, Conv3DBackpropInputV2 receives a filter_sizes instead of the filter tensor.
2. The V2 ops have a data_format attribute, but the V1 ops assume NDHWC.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_operator_helper.h
+++ b/tensorflow/core/common_runtime/dml/dml_operator_helper.h
@@ -107,7 +107,7 @@ class GetOutputShapeFromInputShapeHelper : public ShapeHelper {
  public:
   std::vector<TensorShape> GetOutputShapes(
       OpKernelContext* ctx,
-      const InitializationHelper* initialization_helper) const {
+      const InitializationHelper* initialization_helper) const override {
     const Tensor& input_tensor =
         ctx->input_is_ref(input_tensor_index)
             ? ctx->mutable_input(input_tensor_index, false)
--- a/tensorflow/core/common_runtime/dml/dml_operator_helper.h
+++ b/tensorflow/core/common_runtime/dml/dml_operator_helper.h
@@ -102,12 +102,21 @@ class ScalarOutputShapeHelper : public ShapeHelper {
   };
 };
 
-class GetOutputShapeAsInputShapeHelper : public ShapeHelper {
+template <int input_tensor_index>
+class GetOutputShapeFromInputShapeHelper : public ShapeHelper {
  public:
   std::vector<TensorShape> GetOutputShapes(
       OpKernelContext* ctx,
-      const InitializationHelper* initialization_helper) const override;
+      const InitializationHelper* initialization_helper) const {
+    const Tensor& input_tensor =
+        ctx->input_is_ref(input_tensor_index)
+            ? ctx->mutable_input(input_tensor_index, false)
+            : ctx->input(input_tensor_index);
+    return {input_tensor.shape()};
+  }
 };
+
+using GetOutputShapeAsInputShapeHelper = GetOutputShapeFromInputShapeHelper<0>;
 
 class BroadcastedOutputShapeInitHelper : public InitializationHelper {
  public:
--- a/tensorflow/core/kernels/dml_conv_ops.cc
+++ b/tensorflow/core/kernels/dml_conv_ops.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "tensorflow/core/common_runtime/dml/dml_util.h"
 #include "tensorflow/core/framework/bounds_check.h"
 #include "tensorflow/core/framework/register_types.h"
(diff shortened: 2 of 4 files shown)


tensorflow/core/common_runtime/dml/dml_operator_helper.h
```diff
@@ -102,13 +102,22 @@ class ScalarOutputShapeHelper : public ShapeHelper {
...
 };
 
-class GetOutputShapeAsInputShapeHelper : public ShapeHelper {
+template <int input_tensor_index>
+class GetOutputShapeFromInputShapeHelper : public ShapeHelper {
  public:
   std::vector<TensorShape> GetOutputShapes(
       OpKernelContext* ctx,
-      const InitializationHelper* initialization_helper) const override;
+      const InitializationHelper* initialization_helper) const {
```
adtsai: Why did this drop the `override`?

adtsai (approved):

tensorflow/core/kernels/dml_conv_ops.cc
```diff
@@ -1347,4 +1348,598 @@ class DmlDepthwiseConv2DBackpropInputKernel : public DmlKernel {
...
+    int32 dilation_n = GetTensorDim(dilations, data_format, 'N');
+    int32 dilation_c = GetTensorDim(dilations, data_format, 'C');
+    dilation_d = GetTensorDim(dilations, data_format, '0');
+    dilation_h = GetTensorDim(dilations, data_format, '1');
+    dilation_w = GetTensorDim(dilations, data_format, '2');
+
+    OP_REQUIRES(context, (dilation_n == 1 && dilation_c == 1),
+                errors::InvalidArgument(
+                    "Current implementation does not yet support "
+                    "dilation rates in the batch and depth dimensions."));
```
fdwr, line 1401: Comment says `depth`, but test says `dilation_c`. Guessing you meant `channel` instead.

tensorflow/core/kernels/dml_conv_ops.cc
```diff
@@ -1347,4 +1348,598 @@ class DmlDepthwiseConv2DBackpropInputKernel : public DmlKernel {
...
+
+    const int64 filter_depth = filter.dim_size(0);
+    const int64 filter_height = filter.dim_size(1);
+    const int64 filter_width = filter.dim_size(2);
+    const int64 filter_channels = filter.dim_size(3);
+    const int64 out_channels = filter.dim_size(4);
+
+    OP_REQUIRES(context, in_channels % filter_channels == 0,
+                errors::InvalidArgument(
+                    "Input depth must be evenly divisible by filter depth: ",
```
fdwr, line 1454: Same typo here, channel count, not depth, as the filter `in_depth` can be less than or equal to the input depth.

tensorflow/core/kernels/dml_conv_ops.cc
jstoecker: TF actually refers to the channels as depth in this context, so we want our error messages to be consistent with theirs -- it's not a typo. They don't even name the depth dimension in 3D conv: it's just spatial dim 0, spatial dim 1, spatial dim 2... that's why they can reuse the name depth. :(

For Conv2D:

DirectML Name | TensorFlow Name
-|-
N or Batch|N or Batch
Channels |Depth
Height | Height
Width | Width

For Conv3D:

DirectML Name | TensorFlow Name
-|-
N or Batch|N or Batch
Channels |Depth
Depth | Spatial Dim 0
Height | Spatial Dim 1
Width | Spatial Dim 2

tensorflow/core/kernels/dml_conv_ops.cc
jstoecker: See other comment for same resolution.

tensorflow/core/kernels/dml_conv_ops.cc
```diff
@@ -1347,4 +1348,598 @@ class DmlDepthwiseConv2DBackpropInputKernel : public DmlKernel {
...
+class DmlConv3DBackpropFilterKernel : public DmlKernel {
+ public:
+  using InitHelper = Conv3DGradInitHelper<HasDataFormatAttribute, false>;
+
+  explicit DmlConv3DBackpropFilterKernel(DmlKernelConstruction* ctx,
+                                         const InitHelper* init_helper) {
+    CHECK(ctx->GetInputCount() == 3);
+    CHECK(ctx->GetOutputCount() == 1);
+
+    // 2D conv requires 4D tensors
```
fdwr, line 1848: 🤔 3D conv requires 5d?

PR #169, opened 2021-02-02
Improve _FusedConv2D graph optimization for DML

jeffbloo:
This enables fusion for non-spatial convolution using DirectML, and enables NHWC to NCHW layout transposition for _FusedConv2D.


jstoecker:
Just to make sure: we tested that the perf is actually better when enabling this fusion, right?

jstoecker (approved):

adtsai (approved):
:shipit:

jeffbloo:
> Just to make sure: we tested that the perf is actually better when enabling this fusion, right?

I see slight improvements in the expected models (e.g. VGG16, VGG19, and ICNet).  The best usage pattern is generally to issue fused operators to DirectML, since DirectML internally decomposes operations as is beneficial.

PR #171, opened 2021-02-04
Work around file permission issue with build_pip_package

jstoecker:
The build_pip_package target invokes a bash shell script that, among other things, copies files needed for the package to a temporary directory. On Windows the temp directory is somewhere in the msys2 environment, and we recently started to see strange permission issues with the temp directories created in this environment. The workaround is to prepare package sources in a fixed location on the host OS file system.


fdwr (approved):
Thanks Justinian.

PR #174, opened 2021-02-04
Work around timeout in ID3D12Fence::SetEventOnCompletion on WSL

jeffbloo:
This works around a timeout in ID3D12Fence::SetEventOnCompletion which is specific to the WSL kernel implementation.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_gpu_event.h
+++ b/tensorflow/core/common_runtime/dml/dml_gpu_event.h
@@ -77,8 +77,23 @@ struct DmlGpuEvent {
     WaitForSingleObject(h.get(), INFINITE);
 #else
     // Nullptr event blocks CPU until completion (creates and waits on an event
-    // internally)
-    DML_CHECK_SUCCEEDED(fence->SetEventOnCompletion(fence_value, nullptr));
+    // internally).  This API returns an errur on WSL if the fence value
+    // is not reached after 10 seconds (although this may be changed in future versions 
+    // of the WSL kernel). To work around this, failing calls are re-issued
+    // if they took a minimum length of time (arbitrarily, 1 second). 
+    while (true) {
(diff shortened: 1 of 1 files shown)


tensorflow/core/common_runtime/dml/dml_gpu_event.h
```diff
@@ -77,8 +77,23 @@ struct DmlGpuEvent {
     WaitForSingleObject(h.get(), INFINITE);
 #else
     // Nullptr event blocks CPU until completion (creates and waits on an event
-    // internally)
-    DML_CHECK_SUCCEEDED(fence->SetEventOnCompletion(fence_value, nullptr));
+    // internally).  This API returns an errur on WSL if the fence value
```
jstoecker:
```suggestion
    // internally).  This API returns an error on WSL if the fence value
```

tensorflow/core/common_runtime/dml/dml_gpu_event.h
```diff
@@ -77,8 +77,23 @@ struct DmlGpuEvent {
...
+    // is not reached after 10 seconds (although this may be changed in future versions 
+    // of the WSL kernel). To work around this, failing calls are re-issued
+    // if they took a minimum length of time (arbitrarily, 1 second). 
+    while (true) {
```
jstoecker, line 84: Seems like a max timeout before bailing entirely so would be reasonable... maybe a minute or so.

jstoecker (approved):

tensorflow/core/common_runtime/dml/dml_gpu_event.h
jeffbloo: Neither way is perfect, but as-is I think this provides closer parity with native Windows.  We can revisit and consider a broader timeout policy if needed.

PR #176, opened 2021-02-05
LRN grad kernel

jstoecker:
Adds DML kernel for LRNGrad.

### DIFF
--- a/tensorflow/core/kernels/dml_lrn_op.cc
+++ b/tensorflow/core/kernels/dml_lrn_op.cc
@@ -167,26 +167,6 @@ class LRNGradInitHelper : public InitializationHelper {
   const std::shared_ptr<const Attributes> attr_;
 };
 
-// **********************
-// TODO: REMOVE
-// **********************
-enum DML_PREVIEW_OPERATOR_TYPE {
-  DML_PREVIEW_OPERATOR_FIRST = 0xC0000000,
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/dml_lrn_op.cc
```diff
@@ -128,4 +130,107 @@ TF_CALL_float(REGISTER_KERNEL);
...
+
+    OP_REQUIRES(context, in_grads.dims() == 4 && in_image.dims() == 4,
+                errors::InvalidArgument("inputs must be 4-dimensional"));
+    const int64 batch = in_grads.dim_size(0);
+    const int64 rows = in_grads.dim_size(1);
+    const int64 cols = in_grads.dim_size(2);
+    const int64 depth = in_grads.dim_size(3);
+    OP_REQUIRES(
+        context,
+        in_image.dim_size(0) == batch && in_image.dim_size(1) == rows &&
```
fdwr, line 152: Darn, this awkward wrapping really makes it unnecessarily difficult to read code. Since we're not (realistically) going to merge this with TF 1.5 master anyway, can we use a sane clang format for our subset 😅? I suppose since we're not all nicely partitioned off within a subfolder, like with ORT, we can't easily do so.

fdwr (approved):

PR #179, opened 2021-02-09
Add missing ScatterND validation

PatriceVignola:


tensorflow/core/kernels/dml_resource_scatter_ops.cc
```diff
@@ -49,9 +49,37 @@ static Status ValidateResourceScatter(const Tensor& indices,
   return Status::OK();
 }
 
+static bool ValidEmptyOutputShape(int64 num_inputs, int64 num_indices,
+                                  int64 num_updates) {
+  if (num_indices == 0 && num_updates == 0) {
+    return true;  // regardless of num_inputs ?= 0, covers both cases
+  }
+  // now we want all 3 tensors to have values
+  return (num_inputs != 0 && num_indices != 0 && num_updates != 0);
```
fdwr, line 58: Are negative values permitted in TF? Otherwise semantically it makes much more sense to say the element count is `> 0`. (btw, the naming is confusing, because `num_inputs` sounds like the number of tensor inputs, not the number of tensor elements in the tensor e.g. num_input_elements).

fdwr (approved):

tensorflow/core/kernels/dml_resource_scatter_ops.cc
PatriceVignola: I don't disagree with any of your comments. This was taken as-is from the preexisting scatter_op.cc

jstoecker (approved):

PR #180, opened 2021-02-09
Change the datatype of tensors in debug tests from float64 to float32

PatriceVignola:
The debug tests use np.array, which creates arrays of float64 by default. The specific datatypes for debug tests are completely irrelevant because we don't even compare the values, but since DML doesn't support float64, the tensors won't get placed on the device and the tests will fail.


fdwr (approved):

jstoecker:
Apparently if you use `tf.constant` the default dtype is tf.float32, but if you use `constant_op.constant` the default is inferred from the input data. :\

jstoecker (approved):

adtsai (approved):

PR #182, opened 2021-02-10
Unregister segment ops

PatriceVignola:
Implementing the segment ops with dmlx requires way too much memory in some cases and can become too big for dmlx's temporary buffer which causes some models to crash. If we want to support this op in the future, we should add a new DML API for it. In the meantime, we have to unregister them.


fdwr (approved):
I confess ignorance regarding the cases where it fails, but the change itself looks fine/safe.

jstoecker:
Can you add a comment to the op report metadata that indicates these are not implemented using DMLX for this reason?

jstoecker (approved):

adtsai (approved):
:shipit:

PR #183, opened 2021-02-11
Add InvGrad

PatriceVignola:
InvGrad is literally the same thing as ReciprocalGrad, so only a registration is required.

### DIFF
--- a/tensorflow/core/kernels/dml_cwise_ops.cc
+++ b/tensorflow/core/kernels/dml_cwise_ops.cc
@@ -713,6 +713,8 @@ REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(RsqrtGrad, y * (-0.5f * x) * (x * x),
                                            kNchwDimensionCount)
 REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(ReciprocalGrad, -y* x* x,
                                            kNchwDimensionCount)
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/dml_cwise_ops.cc
```diff
@@ -713,6 +713,8 @@ REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(RsqrtGrad, y * (-0.5f * x) * (x * x),
                                            kNchwDimensionCount)
 REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(ReciprocalGrad, -y* x* x,
                                            kNchwDimensionCount)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(InvGrad, -y* x* x,
```
fdwr, line 716:
```suggestion
REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(InvGrad, -y * x * x,
```

fdwr (approved):

tensorflow/core/kernels/dml_cwise_ops.cc
PatriceVignola: Clang format formats it that way, probably because it's used in a macro and it's missing context, so it assumes they are pointers.

fdwr: (🤦‍♀️ Stupid autoformatters. I wish GitHub had a "sad" reaction.)

PR #186, opened 2021-02-18
Add Rint

PatriceVignola:


fdwr (approved):
Probably better to set the rounding mode field explicitly, but the default 0 happens to work.

jstoecker (approved):

PR #194, opened 2021-03-03
Move initialization binding out of context lock

jstoecker:
This change moves the binding logic for DML operator initialization outside of the DmlExecutionContext lock, like we did earlier for operator execution. This does reduce lock contention, but it won't have a profound effect on performance since initialization occurs once for cached DML kernels. This change _may_ improve performance in models with extremely large numbers of unique ops that overflow the kernel cache, but that is not the primary motivation for this change; rather, this change will enable op execution, op initialization, and device-to-device copies to be batched up and flushed without acquiring the lock for any longer than it takes to append each operation to a batch owned by the execution context.


adtsai (approved):
:shipit:

PR #197, opened 2021-03-05
Batch execution context calls

jstoecker:
This change builds on the previous refactoring changes to further eliminate lock contention in the DML execution context. Recall that the execution context (EC) is the primary mechanism for recording into a D3D command list and submitting it to the hardware. This results in ~45% performance improvement for heavily CPU bound models like Pixel-RNN.

DML kernels and devices may invoke EC methods in a free-threaded manner, but the current implementation simply holds a lock until the work is recorded. Once a fixed number of commands are recorded the EC flushes its current command list to the hardware queue. This approach works fine for most models, but in some extreme cases (e.g.  model with ~100k op executions per epoch) the lock contention starts to add significant overhead. We reduced some of this overhead in previous changes by moving binding logic outside the lock, but the recording work itself is still under the lock.

This change batches the calls to the EC itself (rather than command recorded in the command list) and relies on a background thread to flush the batched calls. Calls are double buffered so that the background thread can record a batch while kernels continue writing into another batch. As a consequence, the arguments passed the EC methods must be kept alive until the function is batch is flushed. This isn't a problem for most arguments, like D3D resources, which are already kept alive using GPU events and the event queue. Other types of args, like barrier structs, are simply copied and kept in the batch. We do not have any cases where the argument data is large enough to make the copies expensive.

A few other minor changes:
- The function batch may be flushed using either a fixed size capacity or if enough time has elapsed since the last flush. This isn't a perfect heuristic, but it seems to work well enough.
- High-frequency calls like op execution no longer emit tracing events by default, since it can add significant overhead when profiling. This behavior can be modified using the TF_DIRECTML_TRACE_LEVEL environment variable.
- Kernels transfer ownership of the binding table for init/exec rather than letting it destroy when the respective kernel context methods go out of scope: this is slightly more efficient than having the EC take a ref on the IUnknown. The binding tables must live until the EC function batch is flushed.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_execution_context.cc
+++ b/tensorflow/core/common_runtime/dml/dml_execution_context.cc
@@ -257,6 +257,7 @@ DmlGpuEvent DmlExecutionContextImpl::ResourceBarrier(
 StatusOr<DmlGpuEvent> DmlExecutionContextImpl::Flush() {
   assert(!closed_);
   DmlTracing::Instance().LogExecutionContextFlush();
+  VLOG(1) << "DML EC IMPL: Begin Flush (" << operations_recorded_in_current_command_list_ << " cmds)";
 
   if (operations_recorded_in_current_command_list_ == 0) {
     // Nothing to flush
@@ -273,7 +274,10 @@ StatusOr<DmlGpuEvent> DmlExecutionContextImpl::Flush() {
     return status_;
   }
 
-  return GetCurrentCompletionEvent();
+  auto event = GetCurrentCompletionEvent();
+  VLOG(1) << "DML EC IMPL: End Flush; completion fv = " << event.fence_value;
+
+  return event;
 }
 
 void DmlExecutionContextImpl::Close() {
@@ -391,6 +395,7 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
 }
 
 void DmlExecutionContext::InvokeBatchedFunctions() {
+  VLOG(1) << "DML EC: InvokeBatchedFunctions (" << shared_state_->batched_functions.size() << "); current fv = " << shared_state_->impl->GetCurrentCompletionEvent().fence_value;
   for (auto& f : shared_state_->batched_functions) {
     f();
   }
@@ -406,7 +411,7 @@ void DmlExecutionContext::InvokeBatchedFunctions() {
     }
 
     if (state->batched_functions.size() >= 25) {
-      VLOG(1) << "DML ThreadProc flush";
+      VLOG(1) << "DML EC ThreadProc: InvokeBatchedFunctions (" << state->batched_functions.size() << "); current fv = " << state->impl->GetCurrentCompletionEvent().fence_value;
       for (auto& f : state->batched_functions) {
         f();
       }
--- a/tensorflow/core/common_runtime/dml/dml_execution_context.cc
+++ b/tensorflow/core/common_runtime/dml/dml_execution_context.cc
@@ -538,18 +538,25 @@ StatusOr<DmlGpuEvent> DmlExecutionContext::InvokeBatchedFunctionsAndExecute() {
 
 /*static*/ void DmlExecutionContext::ThreadProc(
     std::shared_ptr<SharedState> state) {
+
+  auto last_flush = std::chrono::high_resolution_clock::now();
+  
   while (true) {
     std::unique_lock<std::mutex> lock(state->mutex);
     if (state->exit_requested) {
       break;
     }
 
-    if (state->batched_functions.size() >= state->batch_flush_size_) {
+    std::chrono::duration<double> elapsed = std::chrono::high_resolution_clock::now() - last_flush;
+    auto elapsed_ms = elapsed.count() * 1e3;
+
+    if ((state->batched_functions.size() >= state->batch_flush_size_) || (!state->batched_functions.empty() && elapsed_ms >= 1)) {
       for (auto& f : state->batched_functions) {
         f();
       }
       state->batched_functions.clear();
       state->impl->Flush();
+      last_flush = std::chrono::high_resolution_clock::now();
     } else {
       // Wait for new functions
       state->new_function_enqueued.wait(lock);
--- a/tensorflow/core/common_runtime/dml/dml_execution_context.cc
+++ b/tensorflow/core/common_runtime/dml/dml_execution_context.cc
@@ -32,7 +32,8 @@ DmlExecutionContext::DmlExecutionContext(ID3D12Device* d3d_device,
       d3d_device, dml_device, queue, allocator);
 
   int64 batch_flush_size = 0;
-  Status s = ReadInt64FromEnvVar("TF_DIRECTML_BATCH_FLUSH_SIZE", 0, &batch_flush_size);
+  Status s =
+      ReadInt64FromEnvVar("TF_DIRECTML_BATCH_FLUSH_SIZE", 0, &batch_flush_size);
   if (s.ok() && batch_flush_size != 0) {
     shared_state_->batch_flush_size_ = static_cast<uint32_t>(batch_flush_size);
   }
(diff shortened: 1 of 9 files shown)


jstoecker:
As a general comment, I've noticed that the CPU core running the background thread _can_ become saturated in models like Pixel-RNN. A further improvement that I'm considering after this PR is a fork-join on the batch to be flushed: since the batch is already serialized each thread can record a subset of the batch into a separate command list and the lists can be submitted sequentially the by main background thread.

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +421,176 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+    }
+    state->flush_requested = false;
+
+    // Unlock to allow kernels to resume writing to the new write batch.
+    lock.unlock();
+
+    // Invoke the batched functions and submit the work to the GPU.
+    if (flush) {
+      for (auto& f : batch) {
+        f();
```
adtsai, line 585: Is there any particular reason not to execute these as you go along, rather than waiting until the very end when you want to flush? Would that eliminate the need to double-buffer these batches?

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +421,176 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+      }
+      batch.clear();
+      state->impl->Flush();
+      last_flush_time = std::chrono::high_resolution_clock::now();
```
adtsai: As usual, you should probably use std::steady_clock here because std::high_resolution_clock does different things on Linux and Windows.

tensorflow/core/common_runtime/dml/dml_execution_context.h
```diff
@@ -125,78 +132,96 @@ class DmlExecutionContextImpl {
...
 
  private:
-  std::mutex mutex_;
-  std::unique_ptr<DmlExecutionContextImpl> impl_;
+  static constexpr uint32_t default_batch_flush_size = 100;
+  static constexpr uint32_t default_batch_flush_time_us = 1000;
+
+  struct SharedState {
+    std::mutex mutex;
+    DmlGpuEvent next_flush_event;
```
adtsai, line 193: Is this always the same as the impl's CurrentCompletionEvent()? If not, under what circumstances will their fence values differ?

tensorflow/core/common_runtime/dml/dml_execution_context.h
```diff
@@ -125,78 +132,96 @@ class DmlExecutionContextImpl {
...
   DmlExecutionContext(ID3D12Device* d3d12_device, IDMLDevice* dml_device,
                       ID3D12CommandQueue* queue, DmlAllocator* allocator);
 
+  ~DmlExecutionContext();
+
   void Close() {
-    std::unique_lock<std::mutex> lock(mutex_);
-    impl_->Close();
+    std::unique_lock<std::mutex> lock(shared_state_->mutex);
+    shared_state_->impl->Close();
```
adtsai: It seems pretty hairy that this doesn't flush the pending batch(es). Is this ever called by anyone, given that our devices are now global singletons? If not, can we just delete all these Close() methods to avoid this hairiness?

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +421,176 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+
+StatusOr<DmlGpuEvent> DmlExecutionContext::Flush() {
+  std::unique_lock<std::mutex> lock(shared_state_->mutex);
+
+  auto event = shared_state_->next_flush_event;
+  if (shared_state_->WriteBatch().empty()) {
+    --event.fence_value;
+  }
+
+  shared_state_->flush_requested = true;
```
adtsai, line 525: Won't this result in a flush that isn't well ordered with respect to items being added to the batch? That is, the actual flush could occur at any point arbitrarily far in the future (after some other stuff gets added to the batch, for example). Are there any scenarios where this could cause a problem?

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +421,176 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+  absl::InlinedVector<uint8_t, 16> value_copy(value.size());
+  std::copy(value.begin(), value.end(), value_copy.begin());
+
+  shared_state_->WriteBatch().emplace_back(
+      [=, value_copy = std::move(value_copy)]() {
+        shared_state_->impl->FillBufferWithPattern(
+            dst, dst_offset, dst_size_in_bytes, value_copy);
+      });
+
+  return shared_state_->next_flush_event;
```
adtsai, line 451: Are you missing a `notify_all` in this method?

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +421,176 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+    if (state->exit_requested) {
+      break;
+    }
+
+    auto& batch = state->WriteBatch();
+
+    if (batch.empty()) {
+      // Wait for new work to be batched.
+      state->new_function_enqueued.wait(lock);
+      continue;
```
adtsai, line 562: It may be worthwhile to add a comment that this `continue` is necessary to handle spurious wakeups (otherwise to a casual observer it might look superfluous)

tensorflow/core/common_runtime/dml/dml_execution_context.cc
jstoecker: You mean triple buffering, essentially (batch for kernel threads, batch for execution thread, and batched commands within the command list)? Given that background thread can already become saturated as-is I was thinking to have multiple threads do the recording into the command list; this is a little easier if we batch up the functions and then distribute the recording work. It would be possible with a thread pool by round-robin distributing the function calls, but then we'd need a lock on a command list as well.

You can't eliminate the double buffering without also imposing lock contention between the kernel threads and execution thread, which makes a *huge* difference (~25% perf improvement with a single batch to ~50% with double batching).

tensorflow/core/common_runtime/dml/dml_execution_context.h
jstoecker: I'm going to be refactoring this in a future change anyway, but it wouldn't be correct to call CurrentCompletionEvent here. Impl's version returns N if no ops have been recorded in the command list, or N+1 if they have. Since these methods are batching calls the impl will always return N, whereas we know these methods should return N+1 (they are, by definition, adding an op the command list).

Really though this separation between EC/ECImpl needs to be collaped.

tensorflow/core/common_runtime/dml/dml_execution_context.h
jstoecker: Already deleted all of these in a separate branch. It will come with some more refactoring.

tensorflow/core/common_runtime/dml/dml_execution_context.cc
jstoecker: Yep. Also fixed in a pending branch, but it's fairly harmless here: FillBuffer will never be the last call so it can't hurt (there will always be a Device sync -> Flush at the end of a session). I can fix here though.

tensorflow/core/common_runtime/dml/dml_execution_context.cc
jstoecker: It just never ends up mattering. Flush can't be relied on for CPU/GPU synchronization unless you use the returned GPU event, which by definition means you only wait on stuff already in the batch. For GPU/GPU synchronization you'd also need to use fences or barriers, and the implicit assumption is that kernels execute ops in the correct order (or the order just doesn't matter).

The only times we call Flush externally are for readbacks (just to ensure the GPU work commences) or device sync.

tensorflow/core/common_runtime/dml/dml_execution_context.cc
jstoecker: Going to WontFix this for now, since I'm looking at further splitting up the batch with multiple threads as mentioned.

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -372,4 +415,180 @@ void DmlExecutionContextImpl::CloseCommandListAndExecute() {
...
+}
+
+DmlGpuEvent DmlExecutionContext::ResourceBarrier(
+    absl::Span<const D3D12_RESOURCE_BARRIER> barriers) {
+  std::unique_lock<std::mutex> lock(shared_state_->mutex);
+
+  // The caller may not keep the barriers referenced by the span alive for
+  // longer than this function call, so make a copy and transfer ownership to
+  // the lambda.
+  absl::InlinedVector<D3D12_RESOURCE_BARRIER, 4> barriers_copy;
```
jstoecker:
```suggestion
    absl::InlinedVector<D3D12_RESOURCE_BARRIER, 4> barriers_copy(barriers.begin(), barriers.end());
```

adtsai (approved):

PR #200, opened 2021-03-17
Implement (Stateless)RandomUniformInt kernels

jstoecker:
Implements RandomUniformInt and StatelessRandomUniformInt.

### DIFF
--- a/tensorflow/core/kernels/dml_random_ops.cc
+++ b/tensorflow/core/kernels/dml_random_ops.cc
@@ -77,6 +77,32 @@ dml::Expression UniformHalf(dml::Graph& scope, dml::Expression input_state,
   return dml::Reinterpret(result, DML_TENSOR_DATA_TYPE_FLOAT16) - 1.0f;
 }
 
+// Compute a + b where a is a signed type and b is unsigned.
+dml::Expression SignedAdd(dml::Expression a, dml::Expression b) {
+  auto b_div_2 = b / 2;
+  return a + dml::Reinterpret(b_div_2, a.GetOutputDesc().dataType) +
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/dml_random_ops.cc
```diff
@@ -77,6 +77,36 @@ dml::Expression UniformHalf(dml::Graph& scope, dml::Expression input_state,
...
+  auto b_div_2 = b / 2;
+  return a + dml::Reinterpret(b_div_2, a.GetOutputDesc().dataType) +
+         dml::Reinterpret(b - b_div_2, a.GetOutputDesc().dataType);
+}
+
+// Produces a uniform distribution of integers in the range [lo_value,
+// hi_value). See UniformDistribution<Generator, int32> from
+// random_distributions.h. Requires lo_value < hi_value.
+dml::Expression UniformInt(dml::Graph& graph, dml::Expression input_state,
+                           int32_t lo_value, int32_t hi_value,
```
fdwr: I suppose this naming was copied verbatim from the CPU code, and that we wouldn't want to proper English "low" and "high" here. Also, do "lo_value" and "hi_value" directly correspond to "minval" and "maxval"? Seems they do.

fdwr (approved):

tensorflow/core/kernels/dml_random_ops.cc
jstoecker: Yeah, copied from existing code. I suspect people use "lo" and "hi" since it makes the variable lengths the same, which can be aesthetically pleasing, but then min/max also work. Or perhaps it's just to shorten lines of code to avoid ugly clang formatting. Who knows? :)

PR #201, opened 2021-03-18
Enable MaxPool3DGrad kernel

fdwr:
- Enables MaxPool3DGrad kernel.
- Does *not* re-enable tests in pooling_ops_3d_test.py with "@test_util.skip_dml" because the test actually creates a gradient of MaxPool3D (MaxPool3DGrad) and then a gradient of a gradient (MaxPool3DGradGrad), which then fails as DML has no MaxPool3DGradGrad, yielding the misleading and bewildering error "(0) Invalid argument: Default MaxPooling3dGradGradOp only supportsNDHWC on CPU device type". I'm perplexed why we're not seeing the same error for AvgPool3D though, which also lacks a DML GradGrad operator.
- Do I need to create a private nuget of DirectML before enabling this, or will the nightly build select the proper DirectML.dll?

### DIFF
--- a/tensorflow/core/kernels/dml_pooling_ops.cc
+++ b/tensorflow/core/kernels/dml_pooling_ops.cc
@@ -486,7 +486,7 @@ class DmlMaxPoolGradKernel : public DmlKernel {
         ctx, init_helper, tensor_in_shape, 3);
 
     // TF doesn't use dilations, but DML needs default values of 1.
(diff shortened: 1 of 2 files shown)


jstoecker (changes requested):
This change will require a new DML redist, like you guessed. See the changes to tensorflow/workspace.bzl in #190 for an example of how to reference the new package.

jstoecker (approved):

PR #202, opened 2021-03-19
PIX events & markers

jstoecker:
Adds support for PIX events around session runs (CPU event) and operator execution (GPU events). These events make it much nicer to analyze models with PIX timing captures.

Note that CPU events are not part of WinPixEventRuntime's stable ABI, so this change also includes a copy of WinPixEventRuntime.dll in the TF wheels for Windows. PIX events are disabled by default and must be enabled by setting TF_DIRECTML_TRACE_LEVEL.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_execution_context.cc
+++ b/tensorflow/core/common_runtime/dml/dml_execution_context.cc
@@ -211,6 +211,10 @@ D3D12_COMMAND_LIST_TYPE DmlExecutionContext::GetCommandListTypeForQueue()
     std::shared_ptr<DmlCommandList> command_list,
     std::shared_ptr<DmlCommandQueue> command_queue, uint32_t batch_flush_size,
     uint32_t batch_flush_time_us) {
+#if _WIN32
+  SetThreadDescription(GetCurrentThread(), L"TFDML Execution Thread");
+#endif
+
   auto last_flush_time = std::chrono::steady_clock::now();
 
   while (true) {
@@ -252,6 +256,7 @@ D3D12_COMMAND_LIST_TYPE DmlExecutionContext::GetCommandListTypeForQueue()
     lock.unlock();
 
     if (flush) {
+      DmlTracing::Instance().LogExecutionContextFlush();
       // Record the commands into the command list.
       command_list->Open(batch_completion_event);
       for (auto& command : batch) {
@@ -268,7 +273,7 @@ D3D12_COMMAND_LIST_TYPE DmlExecutionContext::GetCommandListTypeForQueue()
 
       ID3D12CommandList* command_lists[] = {command_list->Get()};
       command_queue->ExecuteCommandLists(command_lists);
-      
+
       batch.clear();
       last_flush_time = std::chrono::steady_clock::now();
     }
--- a/tensorflow/core/common_runtime/dml/dml_tracing.cc
+++ b/tensorflow/core/common_runtime/dml/dml_tracing.cc
@@ -4,6 +4,59 @@
 #include <TraceLoggingProvider.h>
 #include <evntrace.h>
 // clang-format on
+
+#define USE_PIX
+#include "pix3.h"
+#include <d3d12.h>
+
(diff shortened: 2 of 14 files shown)


tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -211,6 +211,10 @@ D3D12_COMMAND_LIST_TYPE DmlExecutionContext::GetCommandListTypeForQueue()
     std::shared_ptr<DmlCommandList> command_list,
     std::shared_ptr<DmlCommandQueue> command_queue, uint32_t batch_flush_size,
     uint32_t batch_flush_time_us) {
+#if _WIN32
+  SetThreadDescription(GetCurrentThread(), L"TFDML Execution Thread");
```
fdwr, line 215: Appears this is a dedicated thread owned by DML, and that no other thread will conflict or compete with naming.

tensorflow/core/common_runtime/dml/dml_tracing.cc
```diff
@@ -88,4 +178,24 @@ void DmlTracing::LogKernelCompute(const std::string& op_type,
...
   }
+}
+
+void DmlTracing::LogExecuteOperatorStart(
+    IDMLCompiledOperator* op, ID3D12GraphicsCommandList* command_list) {
+#if _WIN32
+  if (trace_level_ >= All) {
+    std::vector<char> eventName(100);
+    UINT data_size = (UINT)(eventName.size() * sizeof(char));
+    op->GetPrivateData(kPixEventNameId, &data_size, eventName.data());
```
fdwr, line 189: Does DML just truncate if there's not enough space?

tensorflow/core/common_runtime/dml/dml_tracing.h
```diff
@@ -5,18 +5,20 @@
 // Helper for adding tracing events useful in ETW-based perf analysis
 // (GPUView/WPA). No telemetry.
 class DmlTracing {
+ public:
+  // GUID to associate a PIX event name with an IDMLObject (there is an existing
+  // SetName method, but no public GetName method).
+  constexpr static GUID kPixEventNameId = {0x191d7d5, 0x9fe1, 0x47cc, 0xb6,
+                                           0x46,      0xf6,   0x78,   0xb6,
+                                           0x33,      0x2f,   0x96};
+
```
fdwr, line 14: >GUID to associate ... with an IDMLObject
Is this GUID expected to match the GUID of IDMLObject? (which is "c8263aac-9e0c-4a2d-9b8e-007521a3317c")

fdwr (approved):

tensorflow/core/common_runtime/dml/dml_tracing.h
jstoecker: Nope, just a random GUID only used by TFDML. [IDMLObject](https://docs.microsoft.com/en-us/windows/win32/api/directml/nn-directml-idmlobject) (similar to other base DX-style interfaces) supports a few methods for attaching application-defined data to an instance of the interface. This is sometimes convenient since it allows us to stuff arbitrary data into the operator, like a name for the PIX event, without having to store it separately and pass it through each layer of the DML backend as a separate parameter.

Normally we'd call [IDMLObject::SetName](https://docs.microsoft.com/en-us/windows/win32/api/directml/nf-directml-idmlobject-setname) to attach a string/name to the object, which can be used in PIX events that DirectML itself emits, but unfortunately there's no public method to get the name back out of the object.

fdwr: Should we add one?

tensorflow/core/common_runtime/dml/dml_tracing.cc
jstoecker: DML truncates. See [IDMLObject::GetData](https://docs.microsoft.com/en-us/windows/win32/api/directml/nf-directml-idmlobject-getprivatedata). If we passed in data with 128 bytes and tried to read it out into a buffer with only 100 bytes we'd get the first 100 bytes.

PR #207, opened 2021-03-30
Fix RandomUniformInt's DML validation

PatriceVignola:
DML actually does run validation for no-op kernels, but in this case we were never reaching it because the condition was wrong (RandomUniformInit has 3 inputs, not 4).

### DIFF
--- a/tensorflow/core/kernels/dml_random_ops.cc
+++ b/tensorflow/core/kernels/dml_random_ops.cc
@@ -325,7 +325,7 @@ class RandomUniformInitHelper : public InitializationHelper {
     // This init helper is shared for both "RandomUniform" (real types) and
     // "RandomUniformInt" (integral types). The latter has two extra host-memory
     // tensors for the min and max of the output range.
-    if (ctx->num_inputs() == 4) {
+    if (ctx->num_inputs() == 3) {
       const Tensor& minval = ctx->input(1);
       const Tensor& maxval = ctx->input(2);
       OP_REQUIRES(ctx, TensorShapeUtils::IsScalar(minval.shape()),
--- a/tensorflow/python/kernel_tests/random/random_ops_test.py
+++ b/tensorflow/python/kernel_tests/random/random_ops_test.py
@@ -336,7 +336,6 @@ class RandomUniformTest(RandomOpTestCommon):
       self.assertLess(error.max(), 5 * std)
 
   # Check that minval = maxval is fine iff we're producing no numbers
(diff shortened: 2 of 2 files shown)


fdwr:
> ...RandomUniformInt has 3 inputs, not 4...

Are we sure?

Name | Description
-- | --
shape | A Tensor. Must be one of the following types: int32, int64. The shape of the output tensor.
seed | A Tensor. Must be one of the following types: int32, int64. 2 seeds (shape [2]).
minval | A Tensor. Must be one of the following types: int32, int64. Minimum value (inclusive, scalar).
maxval | A Tensor. Must have the same type as minval. Maximum value (exclusive, scalar).

https://www.tensorflow.org/api_docs/python/tf/raw_ops/StatelessRandomUniformInt

[update] Oh wait, maybe this is the right one:

Name | Description
-- | --
shape | A Tensor. Must be one of the following types: int32, int64. The shape of the output tensor.
minval | A Tensor. Must be one of the following types: int32, int64. 0-D. Inclusive lower bound on the generated integers.
maxval | A Tensor. Must have the same type as minval. 0-D. Exclusive upper bound on the generated integers.

https://www.tensorflow.org/api_docs/python/tf/raw_ops/RandomUniformInt

fdwr (approved):

PatriceVignola:
> > ...RandomUniformInt has 3 inputs, not 4...
> 
> Are we sure?
> 
> Name	Description
> shape	A Tensor. Must be one of the following types: int32, int64. The shape of the output tensor.
> seed	A Tensor. Must be one of the following types: int32, int64. 2 seeds (shape [2]).
> minval	A Tensor. Must be one of the following types: int32, int64. Minimum value (inclusive, scalar).
> maxval	A Tensor. Must have the same type as minval. Maximum value (exclusive, scalar).
> https://www.tensorflow.org/api_docs/python/tf/raw_ops/StatelessRandomUniformInt
> 
> [update] Oh wait, maybe this is the right one:
> 
> Name	Description
> shape	A Tensor. Must be one of the following types: int32, int64. The shape of the output tensor.
> minval	A Tensor. Must be one of the following types: int32, int64. 0-D. Inclusive lower bound on the generated integers.
> maxval	A Tensor. Must have the same type as minval. 0-D. Exclusive upper bound on the generated integers.
> https://www.tensorflow.org/api_docs/python/tf/raw_ops/RandomUniformInt

Yep, the seed is an attribute in the C API.

jstoecker:
Oops. :)

jstoecker (approved):

PR #208, opened 2021-04-02
Force memory growth for UMA adapters

jstoecker:
The DML heap allocator, which is based on the GPU BFC allocator, allocates all available GPU memory upfront. This does not work well for UMA adapters, which use shared system memory for their resources, and frequently results an an immediate TDR at startup when using integrated GPUs. This change force-enables the "allow_growth" configuration option on UMA adapters to allocate memory as needed with smaller allocation footprints.

See https://github.com/microsoft/DirectML/issues/95 for an example repro.


adtsai (approved):
:shipit:

PR #212, opened 2021-04-08
Switch to public DX headers

jstoecker:
Now that the DirectX headers and WSL winadapter code are public under a [separate repo](https://github.com/microsoft/DirectX-Headers) we no longer need the initial copy made in TFDML. This change also removes the copy of DirectMLX.h, which is now fetched directly through the DirectML GitHub repo. These changes make it possible to build TFDML without an insider Windows SDK.


tensorflow/core/common_runtime/dml/dml_common.h
```diff
@@ -21,17 +21,18 @@ limitations under the License.
 #include <chrono>
 
 #ifndef _WIN32
+typedef uint64_t UINT64, ULONG_PTR; // TODO: move to dx headers repo
```
PatriceVignola, line 24: Do we have a task/bug number for this?

PatriceVignola (approved):

tensorflow/core/common_runtime/dml/dml_common.h
jstoecker: Nah I just gotta open a PR in Jesse's repo.

PR #214, opened 2021-04-15
Add IDMLDevice::CompileOperator checks to adapter filtering

PatriceVignola:
Adapter filtering right now has a gap where D3D12 device creation will be successful, but some APIs may be missing down the line and eventually make tfdml crash. This change attempts to correct this by making sure that we can at least compile a simple operator before certifying an adapter as DML-compatible.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_device_factory.cc
+++ b/tensorflow/core/common_runtime/dml/dml_device_factory.cc
@@ -70,14 +70,22 @@ static bool IsUmaAdapter(const DmlAdapter& adapter) {
                                         : D3D_FEATURE_LEVEL_11_0;
 
   ComPtr<ID3D12Device> d3d12_device;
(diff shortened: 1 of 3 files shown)


tensorflow/core/common_runtime/dml/dml_device_factory.cc
```diff
@@ -73,11 +73,11 @@ static bool IsUmaAdapter(const DmlAdapter& adapter) {
   DML_CHECK_SUCCEEDED(D3D12CreateDevice(adapter.Impl()->Get(), feature_level,
```
jstoecker: Let's remove the DML_CHECK_SUCCEEDED call when creating the D3D device as well. It really *shouldn't* fail here (we've already created an adapter), but I see no reason to create the risk in the future.

jstoecker (approved):

fdwr (approved):

PR #229, opened 2021-04-30
Skip a few tests that use kernels DML doesn't implement

jstoecker:


tensorflow/python/keras/layers/embeddings_test.py
```diff
@@ -84,6 +84,8 @@ def test_embedding_correctness(self):
     outputs = model.predict(np.array([[0, 1, 0]], dtype='int32'))
     self.assertAllClose(outputs, [[[1, 1], [2, 2], [1, 1]]])
 
+  # DML doesn't implement ResourceSparseApplyAdagrad
```
PatriceVignola, line 87: Do we want to keep track of the tests that we disable because of unsupported operators?

jstoecker: I don't think we should file bugs, if that's what you mean. It should be sufficient to do a search for `skip_dml` or `is_built_with_dml` to find these cases. The responsibility of validating any new kernels is ultimately up to the developer, which should necessarily ensure test cases hit the code that's being added.

PatriceVignola: Sure, as long as we're diligent with it. For example I've stumbled across a lot of "Disabled because DML doesn't support > 5D" tests lately than we should have re-enabled a long time ago, so it's easy to forget sometimes.

PatriceVignola (approved):

PR #231, opened 2021-05-03
Fix indices broadcasting in Cross kernel

PatriceVignola:


tensorflow/core/kernels/dml_cross_op.cc
```diff
@@ -85,6 +85,9 @@ class DmlCrossKernel : public DmlKernel {
     // Generate the indices [-2, -1, 0], which is the same as [1, 2, 0]
     auto indices = dml::Sequence<int32>(scope, -2, 1, {1, 1, 1, 3});
 
+    indices = dml::Reinterpret(indices, input1.GetOutputDesc().sizes,
+                               dml::TensorDimensions({0, 0, 0, 1}));
```
fdwr, line 89: Tis confusing that we use the term `dml::TensorDimensions` when they're not the dimensions of the tensor at all, but rather strides, as it appears like some of the dimensions are 0's. Can we add this to DirectMLX.h in the future?...

```
using TensorDimensions = SmallVector<uint32_t, 4>;
using TensorStrides = SmallVector<uint32_t, 4>; <----
```

https://github.com/microsoft/DirectML/blob/master/Libraries/DirectMLX.h#L251

(non blocking feedback for this PR)

fdwr (approved):

jstoecker (approved):

PR #232, opened 2021-05-03
Add missing ResourceScatterUpdate registration

PatriceVignola:


fdwr (approved):

tensorflow/core/kernels/dml_scatter_update_ops.cc
```diff
@@ -511,4 +511,13 @@ TF_CALL_float(REGISTER_DML_KERNEL);
 #undef REGISTER_RESOURCE_SCATTER_KERNEL
 #undef REGISTER_RESOURCE_SCATTER_KERNEL_INDEX
 
+REGISTER_KERNEL_BUILDER(
```
jstoecker, line 514: DML already registers this kernel for FLOAT/INT32, FLOAT/INT64, VARIANT/INT32, and VARIANT/INT64. See resource_variable_ops.cc. Am I missing something here?

PatriceVignola: This is the bool registration, which is on the GPU unlike int32/variant. And for some reason the GPU only registers it for int32 indices, so we do the same here.

jstoecker: Ah of course, I'm missing the big obvious `<bool>` specialization. Thanks! :D

PR #233, opened 2021-05-05
Fix error when building random ops with empty shapes

PatriceVignola:
minval >= maxval is valid iff the output is empty, which is a no-op. Therefore, we have to stop the validation early and avoid initializing the kernel altogether.


fdwr (approved):
Interesting the existing comment said "Note that we'll never reach this point for empty output", but I didn't see a check for empty output until your change.

jstoecker (approved):

jstoecker:
The comment was copied from TF's CPU kernel but obviously didn't apply to our validation helper. :)

PR #234, opened 2021-05-05
Disable some ndtr tests that produce INF values

PatriceVignola:
For large values of x, DML's implementation of erfc is not precise enough. We always use 1.0 - erf(x), which always gives 0 when x is relatively large. A more precise implementation would be to use fraction approximation like [this implementation from the python math module](https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Modules/mathmodule.c#L517).

### DIFF
--- a/tensorflow/python/kernel_tests/distributions/special_math_test.py
+++ b/tensorflow/python/kernel_tests/distributions/special_math_test.py
@@ -345,8 +345,6 @@ class NdtrGradientTest(test.TestCase):
           rtol=error_spec.rtol,
           atol=error_spec.atol)
 
-  # DML's implementation for erfc is 1.0 - erf(x), which isn't precise enough to
-  # represent large values of x
   @test_util.run_deprecated_v1
   def test_float32(self):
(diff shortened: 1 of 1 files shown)


fdwr:
Hmm, we currently test a normal distribution around mean 0 and standard deviation of 33, which is within tolerance with C's `erf()`. Does this warrant a change to DirectML HLSL too, or are they unnaturally large numbers?

fdwr (approved):

jstoecker (approved):

PatriceVignola:
> Hmm, we currently test a normal distribution around mean 0 and standard deviation of 33, which is within tolerance with C's `erf()`. Does this warrant a change to DirectML HLSL too, or are they unnaturally large numbers?

The problem is with erfc, not erf. All erfc implementations I've seen use `1 - erf(x)` as their default case, but switch to approximations and series for small and big numbers. The numbers are not abnormally large, it's just that our implementation is incomplete.

PR #235, opened 2021-05-05
Skip SparseTensorDenseMatMul test

PatriceVignola:


fdwr (approved):

fdwr:
Should the test use that `@test_util.skip_dml` thing?

PatriceVignola:
> Should the test use that `@test_util.skip_dml` thing?

We could, but still there's already a line to skip CPU, I thought I would add the DML skip the same way.

PR #236, opened 2021-05-05
Improve int64 workaround for negative numbers

PatriceVignola:


tensorflow/core/kernels/dml_addn_op.cc
```diff
@@ -103,9 +94,7 @@ class DmlAddNKernel : public DmlKernel {
 // TODO(b/25387198): A special kernel exists for int32 (see aggregate_ops.cc).
 TF_CALL_float(REGISTER_KERNEL);
 TF_CALL_half(REGISTER_KERNEL);
```
jstoecker, line 96: I assume uint32/uint64 were registered by mistake earlier: I don't see them registered for GPU device. Curious if this caused any problems, though?

tensorflow/core/kernels/dml_broadcast_to_op.cc
```diff
@@ -1,12 +1,9 @@
 /* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 Portions Copyright (c) Microsoft Corporation.
```
jstoecker: Copyright banner shouldn't be modified.

tensorflow/core/kernels/dml_broadcast_to_op.cc
```diff
@@ -113,20 +110,6 @@ class DmlBroadcastToKernel : public DmlKernel {
...
-
-    if (Is64BitIntegerType(output->dtype())) {
-      ctx->ZeroBuffer(ctx->CreateBufferForTensor(*output));
-    }
-
-    return DmlKernel::Compute(ctx);
-  }
 };
 
 #define REGISTER_KERNEL(type)                                         \
```
jstoecker, line 118: I don't think it's necessary to constrain Tidx because the operator only supporrts DT_INT32 and DT_INT64. Probably cleaner to just have a single REGISTER_KERENL_BUILDER macro and remove the type constraint.

tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -481,4 +481,68 @@ DML_SCALAR_UNION ScalarUnion(double value, DML_TENSOR_DATA_TYPE data_type) {
 
   return scalar;
 }
+
+// TFDML #24881131
+Expression ConvertInt32ToInt64(Graph& scope, Expression int32_input,
```
jstoecker, line 486: I think you've taken the crown from RGB2HSV.

jstoecker (approved):

tensorflow/core/kernels/dml_broadcast_to_op.cc
PatriceVignola: I'm super confused, I have no idea why this got modified.

tensorflow/core/kernels/dml_addn_op.cc
PatriceVignola: No problems that I know of, but TF is very careful with what types they register for GPU devices since it can affect operator placement and performance. It's better to just follow whatever CUDA registers until we have a reason to register more types. And supporting uint64 means having one more hack (aka double strides and zero the output), which I don't really want to keep if we don't have to.

tensorflow/core/kernels/dml_broadcast_to_op.cc
PatriceVignola: Sure, this is common to a lot of our operators as well. This PR is already pretty big so I'll keep it this way for now.

tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -481,4 +481,68 @@ DML_SCALAR_UNION ScalarUnion(double value, DML_TENSOR_DATA_TYPE data_type) {
...
+  // Shift the int32 values to the right by 31 to preserve only the sign bit
+  const auto shift =
+      dml::ScalarTensor<uint32_t>(scope, 31, {1, 1, 1, num_elements});
+  const auto sign_bits =
+      dml::Reinterpret(result, DML_TENSOR_DATA_TYPE_UINT32) >> shift;
+
+  // Negate the sign bit to fill all bits with 1's if the value was negative, or
+  // leave them to 0 if it was positive
+  auto complement = -dml::Reinterpret(sign_bits, DML_TENSOR_DATA_TYPE_INT32);
+
```
fdwr, line 510: Clever, composing our missing SAR operator via >> unsigned with -.

tensorflow/core/kernels/dml_concat_op.cc
```diff
@@ -227,30 +227,26 @@ class DmlConcatKernel : public DmlKernel {
...
     auto inputs = GetDmlTensorDescs(tensors.inputs);
-    auto outputs = GetDmlTensorDescs(tensors.outputs);
+    auto scope = dml::Graph(ctx->GetDmlDevice());
 
-    DML_JOIN_OPERATOR_DESC op_specific_desc = {};
-    op_specific_desc.InputCount = inputs.size();
-    op_specific_desc.InputTensors = inputs.data();
-    op_specific_desc.OutputTensor = outputs.data();
-    op_specific_desc.Axis = kNchwDimensionCount - 2;
+    std::vector<dml::Expression> input_tensors;
```
fdwr: (minor) Wasn't there some abseil mini-vector of a min locally allocated size that you used elsewhere?

tensorflow/core/kernels/dml_cwise_ops.cc
```diff
@@ -401,336 +285,248 @@ class DmlUnaryScaleBiasKernel : public DmlKernel {
...
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(SqrtGrad, y * 0.5f / x, 8)
 REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(RsqrtGrad, y*(-0.5f * x) * (x * x),
-                                           kNchwDimensionCount)
-REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(ReciprocalGrad, -y* x* x,
-                                           kNchwDimensionCount)
-REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(InvGrad, -y* x* x,
-                                           kNchwDimensionCount)
+                                           8)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(ReciprocalGrad, -y* x* x, 8)
+REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(InvGrad, -y* x* x, 8)
```
fdwr:
```suggestion
REGISTER_DML_COMPOSITE_BINARY_FLOAT_KERNEL(InvGrad, (-y * x * x), 8)
```

Nit parens to avoid the autoformatter's braindeadness. (ah, I see this already existed)

tensorflow/core/kernels/dml_gather_op.cc
```diff
@@ -319,10 +306,6 @@ class DmlGatherKernel : public DmlKernel {
```
fdwr, line 302: Is the comment above now stale?

// Because we can't guarantee that
// our output tensor's memory is zero'd, we need to do so manually prior to
// running running gather.

 (for some reason, I can't properly anchor the comment to the actual line above?? GitHub... 🤦‍♀️)

tensorflow/core/kernels/dml_reduce_ops.cc
```diff
@@ -171,6 +171,19 @@ class DmlReduceKernel : public DmlKernel {
         reduce_function == DML_REDUCE_FUNCTION_MIN ||
         reduce_function == DML_REDUCE_FUNCTION_MAX;
 
+    // Arg functions in TF are defined with an int64 output, but the indices
+    // cannot be negative we output uint32 values with strides instead of int32
+    // with padding
```
fdwr: "since"

    // Arg functions in TF are defined with an int64 output, but since the indices
    // cannot be negative, we output uint32 values with strides instead of int32
    // with padding.

tensorflow/core/kernels/dml_resource_scatter_ops.cc
```diff
@@ -421,6 +421,10 @@ class DmlResourceScatterNDBinaryKernel : public DmlKernel {
     auto indices = dml::InputTensor(graph, 1, inputs[1]);
     auto updates = dml::InputTensor(graph, 2, inputs[2]);
     auto strides = dml::InputTensor(graph, 3, inputs[3]);
+
+    // TODO: Remove the Is64BitIntegerType hack when DML has a more solid
+    // solution for 64 bit datatypes
+    // TFDML #24881131
     auto result = BinaryOp()(graph, input, indices, updates, strides,
                              Is64BitIntegerType(indices_dtype));
```
fdwr, line 430: Does scatter already work as-is, because I don't see the `result = dml::ConvertInt32ToInt64(scope, result);` that you added elsewhere?

fdwr (approved):

tensorflow/core/kernels/dml_resource_scatter_ops.cc
PatriceVignola: Scatter isn't registered for int64 output since CUDA doesn't register it either. The Is64BitIntegerType here is for the input indices.

PR #239, opened 2021-05-14
Add 8D support for BatchMatMul

PatriceVignola:
DML's GEMM operator doesn't support more than 4D, but we can manually broadcast the batch dimensions using identity and collapse them together, therefore giving us 4D inputs instead of 8D.

### DIFF
--- a/tensorflow/core/kernels/dml_matmul_op.cc
+++ b/tensorflow/core/kernels/dml_matmul_op.cc
@@ -107,25 +107,25 @@ class BaseBatchMatMulInitHelper : public InitializationHelper {
                     "In[0] and In[1] must have compatible batch dimensions: ",
                     in0_shape.DebugString(), " vs. ", in1_shape.DebugString()));
 
-    auto d0 = in0_shape.dim_size(in0_shape.dims() - 2);
-    auto d1 = in0_shape.dim_size(in0_shape.dims() - 1);
-    auto d2 = in1_shape.dim_size(in1_shape.dims() - 2);
-    auto d3 = in1_shape.dim_size(in1_shape.dims() - 1);
+    auto in0_rows = in0_shape.dim_size(in0_shape.dims() - 2);
+    auto in0_cols = in0_shape.dim_size(in0_shape.dims() - 1);
+    auto in1_rows = in1_shape.dim_size(in1_shape.dims() - 2);
+    auto in1_cols = in1_shape.dim_size(in1_shape.dims() - 1);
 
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/dml_matmul_op.cc
```diff
@@ -79,57 +79,86 @@ class BaseBatchMatMulInitHelper : public InitializationHelper {
...
+
+    BCast batches_bcast(BCast::FromShape(in0_batches_shape),
+                        BCast::FromShape(in1_batches_shape));
+
+    OP_REQUIRES(ctx, batches_bcast.IsValid(),
+                errors::InvalidArgument(
+                    "In[0] and In[1] must have compatible batch dimensions: ",
+                    in0_shape.DebugString(), " vs. ", in1_shape.DebugString()));
+
+    auto d0 = in0_shape.dim_size(in0_shape.dims() - 2);
```
fdwr: [minor] For naming, how about `in0w`, `in0h`, `in1w`, `in1h`? Seems more intuitive to me given they represent the matrices' width and height.

tensorflow/core/kernels/dml_matmul_op.cc
```diff
@@ -331,17 +353,46 @@ class DmlBatchMatMulKernel : public DmlKernel {
...
+    auto a_tensor = dml::InputTensor(scope, 0, input_descs[0]);
+    auto b_tensor = dml::InputTensor(scope, 1, input_descs[1]);
+
+    auto a_tensor_sizes = a_tensor.GetOutputDesc().sizes;
+    auto b_tensor_sizes = b_tensor.GetOutputDesc().sizes;
+
+    // DML doesn't support more than 4 dimensions for GEMM, so do the
+    // broadcasting manually by using identity beforehand
+    if (a_tensor_sizes.size() > kNchwDimensionCount) {
+      a_tensor = dml::Identity(a_tensor);
```
fdwr, line 373: Can the identity copy be elided if the leading batch dimensions are all the same between in0 and in1? (like how we internally flatten leading dimensions in DML's `Optimize()` call for many operators)

fdwr (approved):

tensorflow/core/kernels/dml_matmul_op.cc
PatriceVignola: If all leading batch dimensions are the same, then `a_tensor_sizes.size()` will always be `3` because of the broadcasting done in the initialization helper.

PR #240, opened 2021-05-14
Enable outdated disabled tests

PatriceVignola:
I went through all the disabled tests and re-enabled the ones that are now passing


fdwr (approved):
Cool. Glad many of the ND ops could be re-enabled now. (makes me feel like the work was worth it)

jstoecker (approved):

PR #244, opened 2021-05-19
Make DML registrations the biggest union of the types that CUDA registers and that DML supports

PatriceVignola:


tensorflow/core/kernels/dml_cross_op.cc
```diff
@@ -105,6 +105,18 @@ class DmlCrossKernel : public DmlKernel {
     // [y, z, x] -> [z, x, y]
     auto vector4 = dml::GatherElements(vector3, indices, 3);
 
+    const auto dtype = ctx->GetOutputDataType(0);
+    const bool is_low_precision_int = dtype == DT_UINT16 || dtype == DT_INT16 ||
+                                      dtype == DT_UINT8 || dtype == DT_INT8;
+
+    // DML doesn't support less than 32bits precision for multiply and subtract
```
fdwr, line 112: Interesting, usage of int8 for multiply. I wonder how TF handles overflow. Do you know if it just wraps?

tensorflow/core/kernels/dml_matrix_diag_ops.cc
```diff
@@ -184,9 +184,12 @@ class DmlMatrixDiagKernel : public DmlKernel {
       batch_size *= input_shape.dim_size(i);
     }
 
-    auto elem_count_per_batch = static_cast<uint32_t>(input_shape.num_elements() / batch_size);
-    auto output_height = static_cast<uint32_t>(output_shape.dim_size(output_shape.dims() - 2));
-    auto output_width = static_cast<uint32_t>(output_shape.dim_size(output_shape.dims() - 1));
+    auto elem_count_per_batch =
```
fdwr, line 187: [nit] Why though? 🙃 It's more readable on single lines rather than interleaved names and equations.

image

(resolve if this is stupid clangformat related)

tensorflow/core/kernels/dml_one_hot_op.cc
```diff
@@ -152,6 +152,13 @@ class DmlOneHotKernel : public DmlKernel {
     indices_info.desc = DmlTensorDesc::Create(indices_type, indices_shape_dml,
                                               indices_shape_dml);
 
+    // Although DML's OneHot supports int32, the way it handles negative indices
+    // is different from TensorFlow. In negative indices in DML are relative to
+    // the end, but in TensorFlow they are always out of bounds. By forcing the
+    // indices to be unsigned we make sure that they are always out of bounds if
+    // their signed representation is negative.
+    indices_info.desc.ForceUnsignedDataType();
```
fdwr, line 160: It surprises me, because I thought many TF operators used the trend of negative meaning from-the-end.

tensorflow/core/kernels/dml_resize_op.cc
```diff
@@ -127,23 +115,9 @@ class DmlResizeKernel : public DmlKernel {
     DataType tf_input_data_type = ctx->GetInputDataType(0);
     DML_TENSOR_DATA_TYPE dml_input_data_type = result.GetOutputDesc().dataType;
 
-    bool is_identity = input_tensor_shape == output_tensor_shape;
-
-    // ResizeBilinear only supports DT_FLOAT as an output type, and DML doesn't
-    // support mixed data types for Resample. Therefore, we need to cast all
-    // non-float input types to DML_TENSOR_DATA_TYPE_FLOAT32.
-    bool cast_to_float =
-        CastInputToFloat(interpolation_mode, tf_input_data_type, is_identity);
```
fdwr, line 136: Ah, so we're abandoning casting from uint8 to float for now? Yeah, DML doesn't support uint8 directly. (I wonder if transparently casting inside DirectML.dll would be useful, or if it would be better to keep always requiring explicit casting from the caller first. 🤔)

fdwr (approved):

jstoecker (approved):

tensorflow/core/kernels/dml_cross_op.cc
PatriceVignola: I don't know the behavior in TensorFlow, but at least it's consistent. It's unfortunate that the behavior between native vs emulated lower precision support isn't consistent in our case :/

tensorflow/core/kernels/dml_matrix_diag_ops.cc
PatriceVignola: This is stupid clangformat :(

tensorflow/core/kernels/dml_resize_op.cc
PatriceVignola: I realized that CUDA doesn't register those types so we were unnecessarily burdening ourselves with unnecessary hacks.

PR #246, opened 2021-05-19
Disable checkpoint tests and run util tests in float32

jstoecker:
- Disables checkpoint tests (no interaction with DML backend) since they appear to have sporadic permission issues when renaming or deleting files in Windows. Not worth our time to run these in TFDML.
- Use float32 tensors in TraceModelCallTest to avoid op placement issues with numpy default FLOAT64 values

### DIFF
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -5739,6 +5739,9 @@ cuda_py_test(
         "//tensorflow/python/data/ops:dataset_ops",
     ],
     xla_enable_strict_auto_jit = True,
+    # Sporadic permissions issues on Windows related to moving/deleting temp dirs.
+    # This test is unrelated to DML usage.
+    tags = ["no_dml"],
 )
 
 tf_py_test(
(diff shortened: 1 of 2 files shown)


fdwr (approved):

tensorflow/python/BUILD
```diff
@@ -5739,6 +5739,9 @@ cuda_py_test(
         "//tensorflow/python/data/ops:dataset_ops",
     ],
     xla_enable_strict_auto_jit = True,
+    # Sporadic permissions issues on Windows related to moving/deleting temp dirs.
+    # This test is unrelated to DML usage.
+    tags = ["no_dml"],
```
PatriceVignola: Could we use the no_windows tag instead? If it's unrelated to Windows and not specific to DML, we'd want to still run them in Linux (and presumably they would also fail in CPU and GPU builds on Windows).

jstoecker: Same errors even when running upstream version of TF. It has nothing to do with DML. I didn't do "no_windows" to make it easier to look up tests we intentionally disabled in this fork, but since this won't merge I suppose it doesn't matter.

PR #247, opened 2021-05-19
Fix DeepCopy using wrong offsets

PatriceVignola:


tensorflow/core/kernels/dml_deepcopy_op.cc
```diff
@@ -57,9 +57,10 @@ class DmlDeepCopyKernel : public OpKernel {
     constexpr uint64_t src_offset = 0;
```
fdwr: Are these variables even used? Can they be deleted?

tensorflow/core/kernels/dml_deepcopy_op.cc
```diff
@@ -57,9 +57,10 @@ class DmlDeepCopyKernel : public OpKernel {
...
 
     execution_context->CopyBufferRegion(
-        output_buffer.Resource(), dst_offset, D3D12_RESOURCE_STATE_COPY_DEST,
-        input_buffer.Resource(), src_offset, D3D12_RESOURCE_STATE_COPY_SOURCE,
-        input.TotalBytes());
+        output_buffer.Resource(), output_buffer.Offset(),
+        D3D12_RESOURCE_STATE_COPY_DEST, input_buffer.Resource(),
+        input_buffer.Offset(), D3D12_RESOURCE_STATE_COPY_SOURCE,
+        output_buffer.SizeInBytes());
```
fdwr, line 60: Can the output buffer be larger than the input buffer, or vice versa?

fdwr (approved):

fdwr (approved):

tensorflow/core/kernels/dml_deepcopy_op.cc
PatriceVignola: No, because we create the output with the same shape as the input.

PR #250, opened 2021-05-20
Address LSTM/RNN/GRU test issues

jstoecker:
Fixes some test methods in the LSTM/GRU/RNN test classes. These failures fall into three buckets:
1. UnsortedSegmentSum kernel isn't implemented in DML, so Keras' eager execution modes will hit colocation issues when placing random initializer on DML device. We don't want to unregister the random generator op obviously, so for now we skip these tests when run eagerly; they'll still be run in V1 session mode.
2. Qr kernel isn't implemented in DML and Keras' V1 session mode tests will try to force *all* ops on the same device. Fix is to allow fallback to CPU in these tests.
3. Some float64 tests are disabled because DML largely doesn't support float64.

### DIFF
--- a/tensorflow/python/keras/layers/simplernn_test.py
+++ b/tensorflow/python/keras/layers/simplernn_test.py
@@ -43,6 +43,8 @@ class SimpleRNNLayerTest(keras_parameterized.TestCase):
         input_shape=(num_samples, timesteps, embedding_dim))
 
   def test_float64_SimpleRNN(self):
+    if test.is_built_with_rocm:
+      self.skipTest("Double type is yet not supported in DML")
     num_samples = 2
     timesteps = 3
     embedding_dim = 4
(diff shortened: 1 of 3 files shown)


tensorflow/python/keras/layers/simplernn_test.py
```diff
@@ -43,6 +43,8 @@ def test_return_sequences_SimpleRNN(self):
         input_shape=(num_samples, timesteps, embedding_dim))
 
   def test_float64_SimpleRNN(self):
+    if test.is_built_with_rocm:
```
PatriceVignola: Shouldn't this be `is_built_with_dml` and shouldn't this be a function call?

jstoecker: Nothing to see here folks. Move along.

PatriceVignola (approved):

PR #257, opened 2021-05-25
Add StridedSliceAssign and ResourceStridedSliceAssign

PatriceVignola:


tensorflow/core/kernels/dml_strided_slice_op.cc
```diff
@@ -501,4 +504,251 @@ TF_CALL_int8(REGISTER_KERNEL);
...
+  }
+
+  StridedSliceAssignInitHelper(OpKernelContext* ctx,
+                               std::shared_ptr<const Attributes> attr) {
+    DCHECK(ctx->input_is_ref(0) || ctx->input(0).dtype() == DT_RESOURCE);
+
+    if (ctx->input(0).dtype() == DT_RESOURCE) {
+      OP_REQUIRES_OK(
+          ctx, LookupResource(ctx, HandleFromInput(ctx, 0), &input_resource_));
+      input_resource_->mu()->lock_shared();
```
jstoecker, line 538: Should this be unlocked if an exception is thrown in one of the OP_REQUIRES_OK calls below? I suspect TF won't be able to recover regardless, but the destructor won't unlock this if there's a validation error. Might be best to use a cleanup like you did in compute and reset it if we get to the end of the constructor successfully.

jstoecker (approved):

PR #263, opened 2021-06-01
Disable testAtan2SpecialValues for DML

PatriceVignola:


fdwr:
We should probably just make DML have well defined values for <0,0> too. PyTorch may encounter this issue too.

https://github.com/Microsoft/DirectXShaderCompiler/issues/1245

Pat notes that "numpy, c, eigen, and ISO/IEC 60559" all return 0. (maybe I should open a DML bug instead)

fdwr (approved):

jstoecker (approved):

PR #275, opened 2021-06-10
For identity reduce cases, forward input instead

Jamather:
This adds an extra step before kernel creation to check if an operation is a trivial identity. if so, the input can be forwarded to the output instead of performing a copy.

I'm sure there are more places where this could be used but for now it assumes 1 input will be forwarded to 1 output.

### DIFF
--- a/tensorflow/core/kernels/dml_kernel_wrapper.cc
+++ b/tensorflow/core/kernels/dml_kernel_wrapper.cc
@@ -103,6 +103,22 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
       return;
     }
 
+    int inputIndexToForward = -1;
+    if (shared_helper->IsInputForwardable(ctx, output_shapes,
+                                          inputIndexToForward)) {
+      // This should be set if true was returned
+      CHECK(inputIndexToForward != -1);
+      // We assume we are forwarding one input to one output for now.
+      CHECK(ctx->num_outputs() == 1);
+      constexpr int outputIndex = 0;
+      Tensor* output;
+      CHECK(ctx->forward_input_to_output_with_shape(
+          inputIndexToForward, outputIndex, output_shapes[outputIndex],
+          &output));
+
+      return;
+    }
+
     DmlKernelConstruction dml_construction(dml_device, ctx, node_def_.get(),
                                            output_shapes, shared_helper);
 
--- a/tensorflow/core/kernels/dml_kernel_wrapper.cc
+++ b/tensorflow/core/kernels/dml_kernel_wrapper.cc
@@ -107,12 +107,12 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
     forwardIndices.reserve(ctx->num_outputs());
 
     for (int i = 0; i < ctx->num_outputs(); ++i) {
-      int inputIndexToForward = -1;
+      absl::optional<int> inputIndexToForward =
+          shared_helper->GetForwardableInputIndex(ctx, output_shapes, i);
       // If the input is considered forwardable for the op, and the shapes match
-      if (shared_helper->IsOutputForwardable(ctx, output_shapes, i, inputIndexToForward)) {
-        // This should be set if true was returned
-        CHECK(inputIndexToForward != -1);
-        forwardIndices.emplace_back(std::make_pair(inputIndexToForward, i));
+      if (inputIndexToForward) {
+        forwardIndices.emplace_back(
+            std::make_pair(inputIndexToForward.value(), i));
       } else {
         // If not all outputs are forwardable, we will forward nothing and
         // proceed with the kernel normally.
--- a/tensorflow/core/kernels/dml_kernel_wrapper.cc
+++ b/tensorflow/core/kernels/dml_kernel_wrapper.cc
@@ -103,18 +103,34 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
       return;
     }
 
-    int inputIndexToForward = -1;
-    if (shared_helper->IsInputForwardable(ctx, output_shapes,
-                                          inputIndexToForward)) {
-      // This should be set if true was returned
-      CHECK(inputIndexToForward != -1);
-      // We assume we are forwarding one input to one output for now.
-      CHECK(ctx->num_outputs() == 1);
-      constexpr int outputIndex = 0;
-      Tensor* output;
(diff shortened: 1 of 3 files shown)


tensorflow/core/kernels/dml_reduce_ops.cc
```diff
@@ -261,7 +279,8 @@ class DmlReduceKernel : public DmlKernel {
         (reduce_helper.ndims() == 0 ||
          (reduce_helper.ndims() == 1 && !reduce_helper.reduce_first_axis()));
 
-    if (is_identity) {
```
Jamather: Now this case is only necessary if output is 64 bit, since the trivial identity case is handled by the kernel wrapper

tensorflow/core/kernels/dml_kernel_wrapper.cc
```diff
@@ -103,6 +103,22 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
...
+                                          inputIndexToForward)) {
+      // This should be set if true was returned
+      CHECK(inputIndexToForward != -1);
+      // We assume we are forwarding one input to one output for now.
+      CHECK(ctx->num_outputs() == 1);
+      constexpr int outputIndex = 0;
+      Tensor* output;
+      CHECK(ctx->forward_input_to_output_with_shape(
+          inputIndexToForward, outputIndex, output_shapes[outputIndex],
+          &output));
```
PatriceVignola: Is there a reason we do this instead of the more commonly used `set_output` found in most CPU/CUDA kernels? I think using `forward_input_to_output_with_shape` is risky because it can fail for legitimate reasons, whereas `set_output` should only fail if there's not enough memory (which is why it returns void). The kernel has parameters that tell us whether an input can be forwarded to an output or not and `set_output` handles all that heavy lifting for us.

tensorflow/core/kernels/dml_reduce_ops.cc
```diff
@@ -66,6 +66,21 @@ class ReduceInitializationHelper : public InitializationHelper {
     return false;
   }
 
+  bool IsInputForwardable(OpKernelContext* ctx,
+                          absl::Span<const TensorShape> output_shapes,
+                          int& inputIndex) const override {
```
PatriceVignola: Why not have the caller pass the output index? That way we can handle multiple outputs right away and the kernel wrapper just has to call this for every output. Maybe something like `IsOutputForwardable(uint32_t output_index, ...)`?

tensorflow/core/kernels/dml_reduce_ops.cc
PatriceVignola: Why is it different for int64? If both the input and output are int64, and the operation is essentially identity, we should be able to just forward the input. The reason that we needed to convert from int32 to int64 here is that the result of identity would write to an unstrided int32 output that we would need to pad with leading 1's or 0's. But since the input is already a correctly padded int64 buffer, we should be able to reuse it as the output.

tensorflow/core/kernels/dml_kernel_wrapper.cc
jstoecker: The idea is to essentially eliminate the cost of identity reductions by forwarding the tensors rather than dispatching or copying. `set_output` performs a copy and returns a status anyway.

However, it's a good point that we should not simply assert success with `CHECK` because the call to `forward_input_to_output_with_shape` can fail, like Pat says. We should attempt to forward but fall back on failure to doing a copy.

PatriceVignola: That's exactly my point though: `set_output` handles all of that for us. It first tries to forward, but if there's a `Params::kNeverForward` flag, it will fall back to allocating and copying. We don't need to bother with the fallback ourselves.

jstoecker: You're right. I was just reading the stale comments in the header. :)

tensorflow/core/kernels/dml_reduce_ops.cc
Jamather: I made this change but i can't predict how this is going to be used so I imagine it's going to be changed more in the future. Right now it assumes all inputs have to be forwarded but i can definitely imagine cases where some inputs can be forwarded but others still need to be computed.

PatriceVignola: This is why adding an output index and calling this function for every output would be a more comprehensive approach here. We should try to make this as generic as possible. Although this is not a common case, some operators have more than one output and I can imagine cases where we would want to tell the kernel wrapper which one of the outputs we want forwarded.

We will be able to use this helper for many more operators, so having the generic signature right away will avoid the need to refactor all helpers in the future.

tensorflow/core/kernels/dml_reduce_ops.cc
```diff
@@ -66,19 +66,29 @@ class ReduceInitializationHelper : public InitializationHelper {
     return false;
   }
 
-  bool IsInputForwardable(OpKernelContext* ctx,
-                          absl::Span<const TensorShape> output_shapes,
-                          int& inputIndex) const override {
+  bool IsOutputForwardable(OpKernelContext* ctx,
```
PatriceVignola: I think a more modern way to handle this would be to return an `absl::optional<int>` instead, that way you don't need to return a boolean AND also set an integer. This is also less error-prone (implementations can forget to set inputIndexToForward) and the caller doesn't need to check for `inputIndexToForward != -1`. The function could be called `GetForwardableInputIndex` or something similar.

tensorflow/core/kernels/dml_kernel_wrapper.cc
```diff
@@ -103,6 +103,38 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
...
+          shared_helper->GetForwardableInputIndex(ctx, output_shapes, i);
+      // If the input is considered forwardable for the op, and the shapes match
+      if (inputIndexToForward) {
+        forwardIndices.emplace_back(
+            std::make_pair(inputIndexToForward.value(), i));
+      } else {
+        // If not all outputs are forwardable, we will forward nothing and
+        // proceed with the kernel normally.
+        forwardIndices.clear();
+        break;
```
PatriceVignola, line 120: This shouldn't be necessary anymore now that we can mix and match forwardable outputs with non-forwardable outputs.

Edit: Ah nevermind, I see why it's still necessary.

PatriceVignola (approved):

PR #277, opened 2021-06-13
Fix data format ops logic and validation

PatriceVignola:
A few things are going on in this PR:

1. Around 6 months ago, a cherry pick was made from the official TF2 master branch to the 1.15 branch in order to fix potential security issues where the src_format and dst_format attributes for data format ops were not validated. The problem is that this cherry-pick inadvertently also extended the validation from 4D to 5D without also porting the 5D implementation for DataFormatDimMap, thus breaking the CPU and CUDA implementations.
2. Even in TF2, the validation for DataFormatVecPermute is broken because it doesn't validate that the length of src_format and dst_format is equal to the size of the input.
3. In addition to fixing the DML implementations for these ops, this PR aims to also fix the CPU implementation and validation.

### DIFF
--- a/tensorflow/core/kernels/data_format_ops.h
+++ b/tensorflow/core/kernels/data_format_ops.h
@@ -28,24 +28,50 @@ template <typename Device, typename T>
 struct DataFormatDimMap {
   void operator()(const Device& d, typename TTypes<T>::ConstFlat x,
                   typename TTypes<T>::Flat y, const TTypes<int>::Vec dst) {
-    auto zero = x.constant(0);
-    auto one = x.constant(1);
-    auto two = x.constant(2);
+    if (dst.size() == 4) {
+      auto zero = x.constant(0);
+      auto one = x.constant(1);
+      auto two = x.constant(2);
 
-    auto f_zero = x.constant(dst(0));
-    auto f_one = x.constant(dst(1));
-    auto f_two = x.constant(dst(2));
-    auto f_three = x.constant(dst(3));
+      auto f_zero = x.constant(dst(0));
+      auto f_one = x.constant(dst(1));
+      auto f_two = x.constant(dst(2));
+      auto f_three = x.constant(dst(3));
 
-    auto four = x.constant(4);
-    auto x_mod = (x + four) % 4;
+      auto four = x.constant(4);
+      auto x_mod = (x + four) % 4;
 
-    auto is_zero = (x_mod == zero);
-    auto is_one = (x_mod == one);
-    auto is_two = (x_mod == two);
+      auto is_zero = (x_mod == zero);
+      auto is_one = (x_mod == one);
+      auto is_two = (x_mod == two);
 
-    y.device(d) = is_zero.select(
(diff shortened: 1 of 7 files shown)


tensorflow/core/kernels/data_format_ops.h
```diff
@@ -28,30 +28,56 @@ template <typename Device, typename T>
...
-    auto is_two = (x_mod == two);
+      auto is_zero = (x_mod == zero);
+      auto is_one = (x_mod == one);
+      auto is_two = (x_mod == two);
 
-    y.device(d) = is_zero.select(
-        f_zero, is_one.select(f_one, is_two.select(f_two, f_three)));
+      y.device(d) = is_zero.select(
+          f_zero, is_one.select(f_one, is_two.select(f_two, f_three)));
+    } else {
```
fdwr, line 50: Should this be `else if (dst.size() == 5)`, which would be more explicit/clearer?

PatriceVignola: This is not my code, it's copied directly from the TF2 branch (when the validation was cherry-picked to TF 1.15 back from TF2, this code was forgotten)

tensorflow/core/kernels/dml_data_format_dim_map.cc
```diff
@@ -23,21 +23,58 @@ limitations under the License.
...
+// Ensure that `src` and `dst` define a valid permutation.
+// Ops defined in this file assume that user specifies a permutation via two
+// string attributes. This check validates that these attributes properly define
+// it to prevent security vulnerabilities.
+static bool IsValidPermutation(const std::string& src, const std::string& dst) {
+  if (src.size() != dst.size()) {
+    return false;
+  }
+
+  std::map<char, bool> characters;
```
fdwr: Can this just be `std::array<bool, 256>` instead of allocating memory in the `map`? If you do so, be sure to cast the `char` to `uint8_t` before using as an index (some compilers are stupid and treat `char` as a signed value by default, which would be an out of bounds an access).

    for (const uint8_t c : src) {
      if (characters[c]) {
        return false;
      }
      characters[c] = true;
    }

tensorflow/core/kernels/dml_data_format_dim_map.cc
```diff
@@ -71,34 +109,52 @@ class DmlDataFormaDimMapKernel : public DmlKernel {
     // gather operation.
     uint32_t src_dst_mapping_packed = 0;
     uint32_t left_shift = 0;
+    uint8_t additional_src_dst_mapping = 0;
```
fdwr: Suggestion: Although this works, conceptually it would be clearer to me (took a moment to figure out 🤨) to make the `src_dst_mapping_packed` a uint64_t (which simplifies the loop below), and then just `&` or `>>` to access the respective parts of it below, on lines 144 and 154. 🤷‍♀️

fdwr (approved):

fdwr (approved):

PR #280, opened 2021-06-16
Fix StridedSlice float64 not being supported in functional ops tests

PatriceVignola:


fdwr (approved):

fdwr:
@PatriceVignola : Not that it's worth it, but you can probably revert this now that DML 1.7 preview is being used (https://github.com/microsoft/tensorflow-directml/pull/291).

PatriceVignola:
@fdwr I don't think we should enable float64 for DML devices unless we can register it for most operators. The performance cost of copying back and forth between the CPU/GPU would probably be very high and we'd have a lot of device placement and colocation issues.

PR #281, opened 2021-06-16
Disable XLA tests

PatriceVignola:


fdwr:
Guessing this is for Linux, since Windows is already excluded.

fdwr (approved):

PR #285, opened 2021-06-16
Disable int64 and uint64 in bitwise tests that use large numbers

PatriceVignola:


fdwr:
All the bitwise operators with uint64 work with current DirectML.dll? (and any other type just needs a reinterpret-cast to uint8/16/32/64) Can you publish a preview DML package instead of disabling the test?

PR #293, opened 2021-06-27
Fix bad strides for uint64 and int64 Invert

PatriceVignola:
We were doubling the number of elements, but since we didn't change the datatype the strides were still being doubled.


fdwr (approved):
Maybe you can just enable the true 64-bit path now with DML 1.7.

PatriceVignola:
> Maybe you can just enable the true 64-bit path now with DML 1.7.

This is next on my list, this small change will help us get green builds until then.

PR #297, opened 2021-06-30
Add .bazelversion file

PatriceVignola:
This is something that should have been added a long time ago but wasn't there on the official TF 1.15 branch. Basically, this allows working with different bazel repos with tools like bazelisk a lot easier because the right version of bazel is automatically downloaded and executed when "bazel" is called on the command line.


fdwr (approved):
<s>Is the file supposed to be .empty?</s> 🤨 Never mind. :b.

PR #299, opened 2021-07-01
Fix permission errors for RenameFile on Windows

PatriceVignola:
Our tests had sporadic errors due to exceptions thrown from the RenameFile function, and after debugging I found out that deleting the file before overwriting fixes the permission issues when the file is still used somewhere else. This likely also fixes issues encountered by Windows users when they try to run models:

https://stackoverflow.com/questions/62258549/failed-to-rename-access-denied-tensorflow
https://github.com/tensorflow/tensorflow/issues/41380


jstoecker (approved):

fdwr (approved):

tensorflow/core/platform/windows/windows_file_system.cc
```diff
@@ -492,17 +493,38 @@ Status WindowsFileSystem::GetFileSize(const string& fname, uint64* size) {
...
+  WIN32_FIND_DATAW find_file_data;
+  HANDLE target_file_handle =
+      ::FindFirstFileW(ws_translated_target.c_str(), &find_file_data);
+  if (target_file_handle != INVALID_HANDLE_VALUE) {
+    auto file_cleanup = gtl::MakeCleanup(
+        [target_file_handle] { ::FindClose(target_file_handle); });
+
+    if (!::DeleteFileW(ws_translated_target.c_str())) {
+      return IOErrorFromWindowsError(
+          strings::StrCat("Failed to rename: ", src, " to: ", target),
```
fdwr, line 515: "rename" or "delete"? (I suppose in the larger picture, it's really neither a rename nor a delete, but a move)

fdwr (approved):

tensorflow/core/platform/windows/windows_file_system.cc
PatriceVignola: I think we should keep rename here. The fact that we delete the file beforehand is an implementation detail.

PR #302, opened 2021-07-01
Enable reduce infinity tests

PatriceVignola:


fdwr (approved):
Yay, unblocked tests.

PR #303, opened 2021-07-01
Fix failing C API tests by unsharding them

PatriceVignola:
The test helpers should respect Bazel's sharding since some tests need to run in a particular order and in the same process.


fdwr (approved):
Don't really understand it ("shard", a piece of sharp glass), but I get the idea that running out of process leads to nondeterministic ordering of results.

PR #304, opened 2021-07-02
Replace the int32 to int64 workaround with a single cast

PatriceVignola:


tensorflow/core/kernels/dml_concat_op.cc
```diff
@@ -240,7 +240,7 @@ class DmlConcatKernel : public DmlKernel {
 
     // TFDML #24881131
     if (Is64BitSignedIntegerType(ctx->GetOutputDataType(0))) {
-      result = dml::ConvertInt32ToInt64(scope, result);
+      result = dml::ConvertInt32ToInt64(result);
```
fdwr, line 243: >ConvertInt32ToInt64 [](http://example.com/codeflow?start=20&length=19)

Given concat supports int64/uint64 directly, can we just pass int64 to dml::Join instead of the upcast later?

Then you could delete:

```
    const dml::TensorPolicy out_policy =
        Is64BitUnsignedIntegerType(output_dtype)
            ? GetEmulatedInt64TensorPolicy()
            : dml::TensorPolicy::Default();
```

```
if (Is64BitSignedIntegerType(output_dtype)) {
```

Same applies to a few other ops in this CR:

- diag
- cast
- fill
- gather
- gather_nd
- inplace (scatter)
- one_hot
- pack
- roll (identity, gather) fillvaluesequence is used for indices output, so int64 not relevant
- slice
- strided_slice
- tile
- transpose

This is a good next step either way though, even if that's deferred for later.

tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -483,66 +483,7 @@ DML_SCALAR_UNION ScalarUnion(double value, DML_TENSOR_DATA_TYPE data_type) {
 }
 
 // TFDML #24881131
-Expression ConvertInt32ToInt64(Graph& scope, Expression int32_input,
```
fdwr, line 486: >ConvertInt32ToInt64 [](http://example.com/codeflow?start=11&length=19)

Adios. Alas, it was a clever workaround : ). #Resolved

fdwr (approved):
:shipit:

tensorflow/core/kernels/dml_concat_op.cc
PatriceVignola: Agree, I'll do it incrementally.

PR #306, opened 2021-07-06
Avoid grpc race conditions in tests by implementing a port server in portpicker on Windows

PatriceVignola:
Calling portpicker functions without using a port server can lead to race conditions and non-unique ports being picked, which then leads to failures in the tests that need to start grpc server. Since portpicker didn't have Windows implementations for the port server, this change includes changes to portpicker that enable this scenario via named pipes.

Most of the changes under third_party/dml/ci/python_portpicker can be ignored and can instead be compared by looking at the PR against the official portpicker repo: [https://github.com/google/python_portpicker/pull/25](https://github.com/google/python_portpicker/pull/25). If this PR is ever merged into the official portpicker repo and a python package is created, we'll be able to remove that subtree.

### DIFF
--- a/third_party/dml/ci/tests/tfdml_test_runner.py
+++ b/third_party/dml/ci/tests/tfdml_test_runner.py
@@ -145,12 +145,12 @@ def _run_test(exe_path, log_device_placement, shard_index, total_shard_count,
 
 
 def _is_distribute_test(exe_path):
-  temp_path = exe_path
+  remaining_path, tail = os.path.split(exe_path)
 
-  while temp_path != "":
-    temp_path, tail = os.path.split(temp_path)
+  while tail != "":
     if tail == "distribute":
       return True
+    remaining_path, tail = os.path.split(remaining_path)
 
   return False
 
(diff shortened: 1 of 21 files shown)


third_party/dml/ci/tests/tfdml_test_runner.py
```diff
@@ -73,13 +73,18 @@ def _run_test(
...
     env_copy["TF_CPP_LOG_DEVICE_PLACEMENT"] = "1"
 
   if os.name != "nt":
     env_copy["TEST_SRCDIR"] = exe_path + ".runfiles"
 
   if test_framework == "abseil":
+    # Specifying a port server for portpicker is necessary to avoid race
+    # conditions when searching for unused ports. @unittest-portserver is the
+    # sample port server that portpicker provides for test environments.
+    env_copy["PORTSERVER_ADDRESS"] = "@unittest-portserver"
```
jstoecker, line 87: I'm not familiar with portpicker, but I assume it starts this server automatically?

PatriceVignola: The portserver will start the server with `@unittest-portserver` by default, and this environment variable tells portpicker which address to listen to.

jstoecker (approved):

PR #307, opened 2021-07-07
Reduce DML memory limit and allocate persistent resources outside arena

jeffbloo:
These changes are intended to align closer with the GPU device's heuristic, and to help avoid paging which could occur when DirectML operations use persistent memory, such as for indexing optimizations.

### DIFF
--- a/tensorflow/core/kernels/dml_ops_common.cc
+++ b/tensorflow/core/kernels/dml_ops_common.cc
@@ -221,15 +221,15 @@ void DmlKernel::Initialize(DmlKernelConstruction* ctx,
             << " persistent resource for kernel "
             << ctx->GetOpKernelContext()->op_kernel().type_string();
 
-    persistent_resource_ =
-        ctx->AllocateDefaultBuffer(exec_binding_props.PersistentResourceSize);
+   DML_CHECK_SUCCEEDED(ctx->GetD3D12Device()->CreateCommittedResource(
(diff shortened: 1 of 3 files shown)


adtsai (approved):

tensorflow/core/kernels/dml_ops_common.cc
```diff
@@ -221,15 +221,15 @@ void DmlKernel::Initialize(DmlKernelConstruction* ctx,
...
-                    exec_binding_props.PersistentResourceSize, " bytes"));
-
-    persistent_resource_binding_ = persistent_resource_.GetBufferBinding();
+   DML_CHECK_SUCCEEDED(ctx->GetD3D12Device()->CreateCommittedResource(
+       &CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT),
+       D3D12_HEAP_FLAG_NONE,
+       &CD3DX12_RESOURCE_DESC::Buffer(exec_binding_props.PersistentResourceSize, D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS),
+       D3D12_RESOURCE_STATE_UNORDERED_ACCESS,
+       nullptr,
+       IID_PPV_ARGS(&persistent_resource_)));
```
PatriceVignola: Is there a chance that this can return `E_OUTOFMEMORY`? If so, we should handle it the same way as before (with `errors::ResourceExhausted`). Otherwise, you can ignore this comment.

PatriceVignola (approved):

PR #308, opened 2021-07-08
Batch to Space / Space to Batch performance

Jamather:
For 4d cases in the batchtospace/spacetobatch kernels, the new code uses our existing depthtospace/spacetodepth operators ([DMLX PR](https://github.com/microsoft/DirectML/pull/139)) with strides to improve performance. For non 4d, the old implementation is used. 

Additionally, performance is improved by skipping the slice (in batch to space) or pad (in space to batch) if it is not needed.

 I also added an input forwarding check for batchtospace because I noticed there is an identity case built into the kernel. I am not sure if we were hitting that case in AIBench but it's beneficial if we ever do.

[Associated DML changes PR](https://microsoft.visualstudio.com/WindowsAI/_git/WindowsAI/pullrequest/6258833)


tensorflow/core/kernels/dml_space_to_batch_op.cc
```diff
@@ -45,6 +45,11 @@ class BaseSpaceToBatchInitHelper : public InitializationHelper {
     return internal_paddings_;
   }
 
+  bool GetIsPaddingRequired() const {
```
PatriceVignola:
```suggestion
  bool IsPaddingRequired() const {
```

tensorflow/core/kernels/dml_batch_to_space_op.cc
```diff
@@ -316,96 +353,165 @@ class DmlBatchToSpaceKernel : public DmlKernel {
...
 
-    for (int i = 1; i < internal_input_shape.dims(); ++i) {
-      reshaped_sizes.push_back(internal_input_shape.dim_size(i));
-    }
+      // Permute input NCHW->WNCH
+      dml::TensorDimensions input_tensor_sizes = {
+          input_sizes[3], input_sizes[0], input_sizes[1], input_sizes[2]};
+      dml::TensorStrides input_tensor_strides = {
+          input_strides[3], input_strides[0], input_strides[1],
+          input_strides[2]};
```
PatriceVignola, line 373: We already have a way of doing that, which is a bit less error-prone (e.g. look at dml_conv_ops.cc):

```
    // The dimensions of the filter tensor are laid out a little differently
    // than what DML expects
    auto filter_layout = {H, W, C, N};

    // The layout of the input/output tensors is determined by the "data_format"
    // attribute
    auto input_output_layout =
        GetDmlTensorLayout(init_helper->GetDataFormat(), kNchwDimensionCount);
```

tensorflow/core/kernels/dml_batch_to_space_op.cc
```diff
@@ -191,6 +205,29 @@ class BaseBatchToSpaceInitHelper : public InitializationHelper {
...
+  absl::optional<int> GetForwardableInputIndex(
+      OpKernelContext* ctx, absl::Span<const TensorShape> output_shapes,
+      int outputIndex) const override {
+    // For batchtospace, we can only forward input 0 to output 0
+    if (outputIndex != 0) {
+      return {};
+    }
+
+    const Tensor& input =
+        ctx->input_is_ref(0) ? ctx->mutable_input(0, false) : ctx->input(0);
```
PatriceVignola: You don't need the `ctx->input_is_ref(0)` check here since we know that the batch/space ops don't have ref inputs.

tensorflow/core/kernels/dml_batch_to_space_op.cc
```diff
@@ -46,6 +46,20 @@ class BaseBatchToSpaceInitHelper : public InitializationHelper {
 
   TensorShape GetOutputShape() const { return output_shape_; }
   int GetInternalBlockDims() const { return internal_block_dims_; }
+  bool GetIsSliceNeeded() const {
```
PatriceVignola:
```suggestion
  bool IsSliceNeeded() const {
```

tensorflow/core/kernels/dml_batch_to_space_op.cc
```diff
@@ -316,96 +353,165 @@ class DmlBatchToSpaceKernel : public DmlKernel {
...
+            props.totalTensorSizeInBytes = DMLCalcBufferTensorSize(
+                dataType, dimension_count, sizes.data(), props.strides->data());
+            return props;
+          });
+
+      // We want to have special output striding for this one operator so it can
+      // mimic BatchToSpace
+      scope.SetTensorPolicy(out_policy);
+      output_before_slice = dml::DepthToSpace(input, internal_block_sizes[0]);
+      scope.SetTensorPolicy(old_policy);
```
PatriceVignola, line 404: Couldn't this whole permutation logic be avoided by simply using a layout logic similar to what `dml_conv_ops.cc` does? For example:

```
    // The dimensions of the filter tensor are laid out a little differently
    // than what DML expects
    auto filter_layout = {H, W, C, N};

    // The layout of the input/output tensors is determined by the "data_format"
    // attribute
    auto input_output_layout =
        GetDmlTensorLayout(init_helper->GetDataFormat(), kNchwDimensionCount);
```

tensorflow/core/kernels/dml_batch_to_space_op.cc
```diff
@@ -316,96 +353,165 @@ class DmlBatchToSpaceKernel : public DmlKernel {
...
+            props.strides = std::move(output_tensor_strides);
+            props.totalTensorSizeInBytes = DMLCalcBufferTensorSize(
+                dataType, dimension_count, sizes.data(), props.strides->data());
+            return props;
+          });
+
+      // We want to have special output striding for this one operator so it can
+      // mimic BatchToSpace
+      scope.SetTensorPolicy(out_policy);
+      output_before_slice = dml::DepthToSpace(input, internal_block_sizes[0]);
```
PatriceVignola, line 403: Don't forget to update the dmlx SHA when this is fixed :)

tensorflow/core/kernels/dml_batch_to_space_op.cc
Jamather: The naming for Dml::TensorDesc vs DmlTensorDesc is so confusing. 
I like that I can retrieve the Dml::TensorDesc from the expressions because it feels more contained. I think for the input it would be easy but for the output I am not sure if the stride hacking I do is compatible with the logic in DmlTensorDesc::Create. I only reorder the output strides (computed based on reordered output sizes). I leave the output sizes the same.

PatriceVignola (approved):

PR #310, opened 2021-07-12
Update portpicker to the latest pip version

PatriceVignola:
The Windows support for portpicker has now made it to the latest beta portpicker package, so we can remove our internal version of it.


fdwr (approved):
Using the official one rather than our own copy is 👍👍.

PR #312, opened 2021-07-15
Add a crash gatherer to log missing failures

PatriceVignola:
Crashes using the GTEST and ABSL test frameworks instantly kill the process without producing a result XML file. This means that tests that crash won't be counted as failures and erroneously show the test automation as successful. However, we can analyze the output logs to look for any tests that don't have an "end tag" (i.e. PASSED, SKIPPED or FAILED) and assume that a crash happened. This technique won't register all the failures, but this is enough to turn a green test pass into a red one and force us to investigate.


third_party/dml/ci/generate_absl_test_summary.py
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""
+Parses an AbslTest XML log to produce a flattened summary of test results with the following structure:
+{
+    "Counts":
```
fdwr, line 7: [minor] Consider naming "Results", just to be consistent with line 20's name (like how you use "Time" and "Errors" inside and outside the "Tests" block), or "ResultCounts"...

third_party/dml/ci/generate_absl_test_summary.py
```diff
@@ -0,0 +1,241 @@
...
+  path_pattern = re.compile(r"Running '(.+)' with shard \d+\.\.\.")
+
+  with open(log_path, 'r', encoding='utf-16') as test_log:
+    line = test_log.readline()
+
+    current_test_title = None
+    current_test_path = ''
+
+    while line:
+      if line.startswith('[ RUN      ]'):
```
fdwr, line 86: How certain are we that "[ RUN      ]" will always be 12 characters? Should it be a pattern instead, like "[ *RUN *]"? (or "[ +RUN +]")

fdwr (approved):

third_party/dml/ci/generate_absl_test_summary.py
PatriceVignola: It's always been 12 characters (if it ever changes then it most likely won't be named "RUN" anymore)

third_party/dml/ci/generate_absl_test_summary.py
PatriceVignola: This requires changing a few other scripts as well. I'll keep it that way for now just to stay consistent with the DML scripts as well.

fdwr: Okie dokie.

third_party/dml/ci/generate_absl_test_summary.py
fdwr: gtest v1.12.0 "[ SEE SPOT RUN ]" 😅

PR #314, opened 2021-07-21
Support tiled resources in primary DML allocator

jstoecker:
Ideally this change would be split into smaller ones, so I apologize in advanced for the large size. 

The major changes:

1. The primary DML suballocator (`D3D12HeapAllocator`) now supports allocations that are backed by multiple heaps (tiling) to better utilize local video memory and avoid demotion to non-local memory. The previous behavior of using contiguous heaps and placed resources is retained as a fallback for hardware that doesn't support tiled buffers.

2. The `D3D12HeapAllocator` no longer maintains a pool of resources for each allocation; instead, it has three otherwise identical resources that remain in fixed states: UAV, COPY_SRC, and COPY_DST. This change is important for change No. 1 because while placed resources are cheap, reserved resources can consume significant memory with their unique page tables. Consequently, `D3D12BufferRegion`s no longer take ownership of pooled resources and are now lightweight wrappers that help manage access to memory regions without kernels having to deal with resource state transitions directly.

3. Significant refactoring of the DML kernel context to move various helpers to a single interface (`DmlDeviceContext`). We previously had kernels calling into DmlKernelContext, DmlKernelConstruction, dml_util, DmlExecutionContext, DmlDeviceContext, and the upload/readback pooled allocators. This makes is extremely tedious to refactor any code involving allocations or copies without touching dozens of files. Going forward kernel code should go straight to the `DmlDeviceContext` for anything related to copies or allocation. It's important to transition most code away from using raw ID3D12Resource* objects and offsets and instead using `D3D12BufferRegion` now that resources aren't pooled: it would be easy to use the wrong ID3D12Resource* handle wrapped by this helper.

4. Memory growth is turned on by default for *all* adapters, not just UMA adapters. It's possible to go back to the GPU device policy by setting the TF_GPU_FORCE_ALLOW_GROWTH environment variable to "false". Memory growth isn't crucial now with reserved resources, but TFDML is most likely used for local training where we shouldn't assume TF deserves all the memory.

A bit more context for change No. 2. Recall that the BFC allocator carves up contiguous allocations and ensures distinct tensors live in physically separate chunks. However, the D3D API exposes GPU memory *primarily* through resources, not virtual addresses, so we must have resource objects to bind memory when executing operators. Creating resources on the fly is expensive, but caching resources is difficult since chunks are continually merged and split: the offsets and sizes of memory with an allocation/heap will vary significantly from resource to resource. The resource pool solved this by having all resources span the entirety of a heap and rely on offsets during binding. Multiple resources (all of which span the same region of heap memory) allowed callers to get unique instances for the purpose of resource barriers.

**NOTE**: this change will require some changes to DML debug layer validation (simultaneous bindings of same resource), which isn't enabled in TF by default.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
+++ b/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
@@ -31,29 +31,35 @@ D3D12BufferRegion::D3D12BufferRegion(
       resource_(std::move(resource)),
       resource_copy_src_state_(std::move(resource_copy_src_state)),
       resource_copy_dst_state_(std::move(resource_copy_dst_state)) {
-  assert(resource_ != nullptr);
-  assert(size_in_bytes_ != 0);
+
+  CHECK(resource_ != nullptr);
+  CHECK(size_in_bytes_ != 0);
+
+  uint64_t buffer_size = resource->GetDesc().Width;
+  CHECK(offset_ < buffer_size);
+  CHECK(size_in_bytes_ <= buffer_size - offset);
+
   assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
-  assert(
-      !resource_copy_src_state ||
-      (resource_copy_src_state->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER &&
-       resource_copy_src_state->GetDesc().Width == resource->GetDesc().Width));
-  assert(
-      !resource_copy_dst_state ||
-      (resource_copy_dst_state->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER &&
-       resource_copy_dst_state->GetDesc().Width == resource->GetDesc().Width));
+  assert(!resource_copy_src_state ||
+         (resource_copy_src_state->GetDesc().Dimension ==
+              D3D12_RESOURCE_DIMENSION_BUFFER &&
+          resource_copy_src_state->GetDesc().Width == buffer_size));
+  assert(!resource_copy_dst_state ||
+         (resource_copy_dst_state->GetDesc().Dimension ==
+              D3D12_RESOURCE_DIMENSION_BUFFER &&
+          resource_copy_dst_state->GetDesc().Width == buffer_size));
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInFixedState() const {
-  return resource_ ? resource_.Get() : nullptr;
+  return resource_.Get();
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInCopySrcState() const {
-  return resource_copy_src_state_ ? resource_copy_src_state_.Get() : nullptr;
+  return resource_copy_src_state_.Get();
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInCopyDstState() const {
-  return resource_copy_dst_state_ ? resource_copy_dst_state_.Get() : nullptr;
+  return resource_copy_dst_state_.Get();
 }
 
 uint64_t D3D12BufferRegion::Offset() const { return resource_ ? offset_ : 0; }
--- a/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
+++ b/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
@@ -31,7 +31,6 @@ D3D12BufferRegion::D3D12BufferRegion(
       resource_(std::move(resource)),
       resource_copy_src_state_(std::move(resource_copy_src_state)),
       resource_copy_dst_state_(std::move(resource_copy_dst_state)) {
-
   CHECK(resource_ != nullptr);
   CHECK(size_in_bytes_ != 0);
 
--- a/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
+++ b/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
@@ -34,19 +34,19 @@ D3D12BufferRegion::D3D12BufferRegion(
   CHECK(resource_ != nullptr);
   CHECK(size_in_bytes_ != 0);
 
-  uint64_t buffer_size = resource->GetDesc().Width;
+  uint64_t buffer_size = resource_->GetDesc().Width;
   CHECK(offset_ < buffer_size);
   CHECK(size_in_bytes_ <= buffer_size - offset);
 
   assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
-  assert(!resource_copy_src_state ||
-         (resource_copy_src_state->GetDesc().Dimension ==
+  assert(!resource_copy_src_state_ ||
+         (resource_copy_src_state_->GetDesc().Dimension ==
               D3D12_RESOURCE_DIMENSION_BUFFER &&
-          resource_copy_src_state->GetDesc().Width == buffer_size));
-  assert(!resource_copy_dst_state ||
-         (resource_copy_dst_state->GetDesc().Dimension ==
+          resource_copy_src_state_->GetDesc().Width == buffer_size));
+  assert(!resource_copy_dst_state_ ||
+         (resource_copy_dst_state_->GetDesc().Dimension ==
               D3D12_RESOURCE_DIMENSION_BUFFER &&
-          resource_copy_dst_state->GetDesc().Width == buffer_size));
+          resource_copy_dst_state_->GetDesc().Width == buffer_size));
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInFixedState() const {
--- a/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
+++ b/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
@@ -34,7 +34,14 @@ D3D12BufferRegion::D3D12BufferRegion(
   assert(resource_ != nullptr);
   assert(size_in_bytes_ != 0);
   assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
-  // TODO: if other resources provided they should match size
+  assert(
+      !resource_copy_src_state ||
+      (resource_copy_src_state->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER &&
+       resource_copy_src_state->GetDesc().Width == resource->GetDesc().Width));
+  assert(
+      !resource_copy_dst_state ||
+      (resource_copy_dst_state->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER &&
+       resource_copy_dst_state->GetDesc().Width == resource->GetDesc().Width));
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInFixedState() const {
--- a/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
+++ b/tensorflow/core/common_runtime/dml/dml_buffer_region.cc
@@ -21,24 +21,40 @@ namespace tensorflow {
 
 D3D12BufferRegion::D3D12BufferRegion(
     uint64_t offset, uint64_t size_in_bytes,
-    D3D12_RESOURCE_STATES resource_state,
-    Microsoft::WRL::ComPtr<ID3D12Resource> resource,
+    Microsoft::WRL::ComPtr<ID3D12Resource> resource_uav_state,
     Microsoft::WRL::ComPtr<ID3D12Resource> resource_copy_src_state,
     Microsoft::WRL::ComPtr<ID3D12Resource> resource_copy_dst_state)
     : offset_(offset),
       size_in_bytes_(size_in_bytes),
-      resource_state_(resource_state),
-      resource_(std::move(resource)),
+      resource_uav_state_(std::move(resource_uav_state)),
       resource_copy_src_state_(std::move(resource_copy_src_state)),
       resource_copy_dst_state_(std::move(resource_copy_dst_state)) {
-  CHECK(resource_ != nullptr);
+  // Get a raw pointer to the first non-null resource passed in. At least one
+  // resource must be provided.
+  first_valid_resource_ = resource_uav_state_.Get();
+  if (!first_valid_resource_) {
+    first_valid_resource_ = resource_copy_src_state.Get();
+  }
+  if (!first_valid_resource_) {
+    first_valid_resource_ = resource_copy_dst_state.Get();
+  }
+  CHECK(first_valid_resource_ != nullptr);
+
+  // Regions cannot be empty.
   CHECK(size_in_bytes_ != 0);
 
-  uint64_t buffer_size = resource_->GetDesc().Width;
+  // Regions cannot extend beyond the size of the resource.
+  uint64_t buffer_size = first_valid_resource_->GetDesc().Width;
   CHECK(offset_ < buffer_size);
   CHECK(size_in_bytes_ <= buffer_size - offset);
 
-  assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
+  // All three resources, if provided, must be identical aside from state.
+  assert(first_valid_resource_->GetDesc().Dimension ==
+         D3D12_RESOURCE_DIMENSION_BUFFER);
+  assert(!resource_uav_state ||
+         (resource_uav_state->GetDesc().Dimension ==
+              D3D12_RESOURCE_DIMENSION_BUFFER &&
+          resource_uav_state->GetDesc().Width == buffer_size));
   assert(!resource_copy_src_state_ ||
          (resource_copy_src_state_->GetDesc().Dimension ==
               D3D12_RESOURCE_DIMENSION_BUFFER &&
@@ -49,8 +65,8 @@ D3D12BufferRegion::D3D12BufferRegion(
           resource_copy_dst_state_->GetDesc().Width == buffer_size));
 }
 
-ID3D12Resource* D3D12BufferRegion::ResourceInFixedState() const {
-  return resource_.Get();
+ID3D12Resource* D3D12BufferRegion::ResourceInUavState() const {
+  return resource_uav_state_.Get();
 }
 
 ID3D12Resource* D3D12BufferRegion::ResourceInCopySrcState() const {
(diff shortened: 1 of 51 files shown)


tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
```diff
@@ -113,45 +301,10 @@ D3D12BufferRegion D3D12HeapAllocator::CreateBufferRegion(
...
-  Allocation* allocation = &it->second;
-
-  // If the sizes don't match, then it means this resource didn't come from this
-  // allocator...
-  assert(allocation->heap->GetDesc().SizeInBytes == resource->GetDesc().Width);
-
-  // Return the resource to the pool
-  allocation->placed_resource_pool.push_back(std::move(resource));
+  return D3D12BufferRegion(
+      tagged_ptr.offset, size_in_bytes, D3D12_RESOURCE_STATE_UNORDERED_ACCESS,
```
jstoecker: No practical effect at the moment, but switch to `initial_state_` to be consistent.

tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
```diff
@@ -17,40 +17,232 @@ limitations under the License.
...
       CD3DX12_HEAP_DESC(size_in_bytes, heap_properties_, 0, heap_flags_);
+  HRESULT create_heap_hr =
+      device_->CreateHeap(&heap_desc, IID_PPV_ARGS(&allocation.heaps.front()));
+  if (dml_util::HrIsOutOfMemory(create_heap_hr)) {
+    LOG(WARNING) << "DML allocator out of memory!";
+    return absl::nullopt;
+  }
+  DML_CHECK_SUCCEEDED(create_heap_hr);
+
+  // Initialize the pool with a single placed resource.
```
jstoecker: Not a pool anymore. Update comment.

tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
```diff
@@ -17,40 +17,232 @@ limitations under the License.
...
+}
 
-void* D3D12HeapAllocator::Alloc(uint64_t size_in_bytes) {
-  if (size_in_bytes == 0) {
-    return nullptr;
+absl::optional<D3D12HeapAllocator::Allocation>
+D3D12HeapAllocator::TryCreateTiledAllocation(uint64_t size_in_bytes) {
+  Allocation allocation = {};
+
+  // Initialize the pool with a single reserved resource. The resource may be
```
jstoecker: Not a pool anymore. Update comment.

tensorflow/core/common_runtime/dml/dml_buffer_region.cc
```diff
@@ -20,23 +20,35 @@ limitations under the License.
...
+  // TODO: if other resources provided they should match size
 }
 
-ID3D12Resource* D3D12BufferRegion::Resource() const {
+ID3D12Resource* D3D12BufferRegion::ResourceInFixedState() const {
   return resource_ ? resource_.Get() : nullptr;
 }
 
+ID3D12Resource* D3D12BufferRegion::ResourceInCopySrcState() const {
+  return resource_copy_src_state_ ? resource_copy_src_state_.Get() : nullptr;
```
fdwr: [minor] Is not equivalent to just say `return resource_copy_src_state_.Get()`? If it's false, then the ComPtr would be nullptr anyway.

Just skimmed it. This is definitely Adrian and Jeff's domain.

tensorflow/core/kernels/dml_zeros_like_op.cc
```diff
@@ -39,8 +39,7 @@ static void SetTensorToZero(OpKernelContext* ctx, const Tensor& tensor) {
   uint8_t pattern[] = {0};
```
jstoecker: Use ZeroBuffer

tensorflow/core/kernels/dml_inplace_ops.cc
```diff
@@ -158,11 +160,9 @@ class DmlInplaceKernel : public DmlKernel {
       return status_or_gpu_event;
     }
 
-    ctx->CopyBufferToBuffer(input_buffer.Resource(), input_buffer.Offset(),
-                            output_buffer.Resource(), output_buffer.Offset(),
-                            input_buffer.SizeInBytes());
+    device_context->CopyBufferToBuffer(input_buffer, output_buffer);
```
jstoecker: Check output_buffer size is identical to input_buffer size to be consistent with original.

tensorflow/core/kernels/dml_parallel_concat_ops.cc
```diff
@@ -83,12 +83,10 @@ class DmlParallelConcatUpdateKernel : public OpKernel {
     // Guard the row index range
     const int64 row_index = (loc_ % nrows + nrows) % nrows;
     const uint64_t dst_offset = output_buffer.Offset() + row_index * stride;
```
jstoecker: adding output offset twice now with subregion call. remove output_buffer.Offset()

tensorflow/core/kernels/dml_tensor_array.cc
```diff
@@ -89,15 +92,11 @@ void DmlConcatTensors(OpKernelContext* ctx, Tensor* output_tensor,
...
         dml_util::CreateBufferForTensor(device, input_tensor);
 
-    // GPU resources are always kept in UAV state
-    const auto barrier_state = D3D12_RESOURCE_STATE_UNORDERED_ACCESS;
-
-    device->GetExecutionContext()->CopyBufferRegion(
-        dst.Resource(), dst_offset, barrier_state, src.Resource(), src.Offset(),
-        barrier_state, bytes_to_copy);
+    device_context->CopyBufferToBuffer(
+        dst, src.Subregion(0, bytes_to_copy));
```
jstoecker: should be applying updated dst_offset to subregion of dst each iteration

tensorflow/core/common_runtime/dml/dml_buffer_region.h
```diff
@@ -24,36 +24,51 @@ class D3D12HeapAllocator;
...
+  D3D12_RESOURCE_STATES ResourceState() const;
 
   DML_BUFFER_BINDING GetBufferBinding() const;
 
   explicit operator bool() const { return resource_ != nullptr; }
 
+  // Creates a subregion at an offset from the start of this region. If no size is provided
+  // the region runs to the end of the current region.
+  inline D3D12BufferRegion Subregion(uint64_t offset, uint64_t size_in_bytes = 0) const {
+    size_in_bytes = size_in_bytes == 0 ? size_in_bytes_ - offset : size_in_bytes;
```
jstoecker: should check offset < size_in_bytes

tensorflow/core/common_runtime/dml/dml_buffer_region.cc
```diff
@@ -20,23 +20,35 @@ limitations under the License.
...
+      resource_state_(resource_state),
       resource_(std::move(resource)),
-      offset_(offset),
-      size_in_bytes_(size_in_bytes) {
+      resource_copy_src_state_(std::move(resource_copy_src_state)),
+      resource_copy_dst_state_(std::move(resource_copy_dst_state)) {
   assert(resource_ != nullptr);
   assert(size_in_bytes_ != 0);
   assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
+  // TODO: if other resources provided they should match size
```
jstoecker: implement todo

tensorflow/core/common_runtime/dml/dml_buffer_region.cc
```diff
@@ -20,23 +20,42 @@ limitations under the License.
...
+      resource_state_(resource_state),
       resource_(std::move(resource)),
-      offset_(offset),
-      size_in_bytes_(size_in_bytes) {
+      resource_copy_src_state_(std::move(resource_copy_src_state)),
+      resource_copy_dst_state_(std::move(resource_copy_dst_state)) {
   assert(resource_ != nullptr);
   assert(size_in_bytes_ != 0);
   assert(resource_->GetDesc().Dimension == D3D12_RESOURCE_DIMENSION_BUFFER);
+  assert(
```
PatriceVignola: nit: this file doesn't look clang formatted anymore

tensorflow/core/common_runtime/dml/dml_buffer_region.h
```diff
@@ -24,36 +24,53 @@ class D3D12HeapAllocator;
...
   uint64_t SizeInBytes() const;
+  D3D12_RESOURCE_STATES ResourceState() const;
 
   DML_BUFFER_BINDING GetBufferBinding() const;
 
   explicit operator bool() const { return resource_ != nullptr; }
 
+  // Creates a subregion at an offset from the start of this region. If no size is provided
+  // the region runs to the end of the current region.
+  inline D3D12BufferRegion Subregion(uint64_t offset, uint64_t size_in_bytes = 0) const {
```
PatriceVignola: nit: format and many other files (probably due to a mass rename). you could probably just run the clang format tool in the `tensorflow/core/common_runtime/dml` folder

tensorflow/core/common_runtime/dml/dml_device_context.cc
```diff
@@ -148,4 +121,151 @@ void DMLDeviceContext::CopyDeviceTensorToCPU(const Tensor* device_tensor,
   event_queue_->Enqueue(status_or_event.ConsumeValueOrDie(), callback);
 }
 
+// TODO: consider moving this code into EC outside of lock
```
PatriceVignola: This TODO could easily get lost. If we don't want to create a tracking bug/task for it, maybe we should at least make it `TODO (tfdml): ...` (same for the one below)

tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
```diff
@@ -17,40 +17,233 @@ limitations under the License.
...
+    }
+  }
+
+  return false;
+}
+
+static uint64_t GetMaxHeapSizeInTiles() {
+  int64 override_value = 0;
+  Status s =
+      ReadInt64FromEnvVar("TF_DIRECTML_MAX_HEAP_TILES", 0, &override_value);
```
PatriceVignola, line 57: This is more of a general comment, but we might consider putting the environment variables specific to tfdml in the README somewhere.

tensorflow/core/common_runtime/dml/dml_device_context.cc
jstoecker: This change has enough refactoring as is so I'm just going to delete the TODO. There's no practical benefit to this anyway.

tensorflow/core/common_runtime/dml/dml_buffer_region.cc
jstoecker: Makes sense. Just copying what was here before. :)

tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
```diff
@@ -17,40 +17,233 @@ limitations under the License.
...
+
+  assert(unmapped_resource_tiles == 0);
+
+  // Copy tile mappings from the primary resource.
+  {
+    D3D12_TILED_RESOURCE_COORDINATE resource_start_coordinate = {};
+    D3D12_TILE_REGION_SIZE resource_region_size = {};
+    resource_region_size.NumTiles = resource_size_in_tiles;
+
+    queue_->CopyTileMappings(
```
adtsai, line 175: It may be worthwhile to add a comment why we don't need to wait for this to complete before returning, because that might not be immediately obvious.

tensorflow/core/common_runtime/dml/dml_buffer_region.h
```diff
@@ -24,36 +24,58 @@ class D3D12HeapAllocator;
...
+    return D3D12BufferRegion(offset_ + offset, size_in_bytes, resource_state_,
+                             resource_, resource_copy_src_state_,
+                             resource_copy_dst_state_);
+  }
+
  private:
-  D3D12HeapAllocator* allocator_ = nullptr;
-  uint32_t allocation_id_ = 0;
   Microsoft::WRL::ComPtr<ID3D12Resource> resource_;
+  Microsoft::WRL::ComPtr<ID3D12Resource> resource_copy_src_state_;
```
adtsai: Strong refs were used here because of the way the pool of placed resources worked: the call to CreateBufferRegion would *transfer* ownership of the ID3D12Resource into the D3D12BufferRegion. When the D3D12BufferRegion was destructed, it would then transfer ownership back to the allocator. Since it appears the allocator now holds exclusive ownership of the 3 resources, you should be able to get rid of the ComPtr's here.

tensorflow/core/common_runtime/dml/dml_device_context.cc
```diff
@@ -35,19 +35,13 @@ void DMLDeviceContext::CopyCPUTensorToDevice(const Tensor* cpu_tensor,
...
   const void* src_data = DMAHelper::base(cpu_tensor);
 
-  D3D12BufferRegion dst = dml_util::CreateBufferForTensor(
-      static_cast<DmlDevice*>(device), *device_tensor);
-  ID3D12Resource* dst_data = dst.Resource();
-  const uint64_t dst_offset = dst.Offset();
-  const auto dst_state =
-      D3D12_RESOURCE_STATE_UNORDERED_ACCESS;  // GPU resources are always kept
-                                              // in UAV state
+  D3D12BufferRegion dst = CreateBufferForTensor(*device_tensor);
```
adtsai: It might actually be worthwhile to rename this to "GetBufferForTensor". It used to be called "Create" because the logical lifetime of the buffers was determined by the returned D3D12BufferRegion. As far as the caller was concerned, CreateBufferForTensor really did create a fresh placed resource that disappeared once the D3D12BufferRegion went away (the fact that they were really cached in a pool was an implementation detail). If the new lifetime semantics are intended to be different, a rename of this function might be warranted.

tensorflow/core/common_runtime/dml/dml_device_context.h
```diff
@@ -52,6 +64,87 @@ class DMLDeviceContext : public DeviceContext {
...
+      _In_opt_ const DML_BUFFER_BINDING* temporary_resource_binding,
+      _In_opt_ const DML_BUFFER_BINDING* persistent_resource_binding);
+
+  // Executes a DML operator. Note that this merely queues the execution; the
+  // returned event will enter the signaled state when it completes.
+  DmlGpuEvent BindAndExecuteOperator(
+      IDMLCompiledOperator* op,
+      Microsoft::WRL::ComPtr<IDMLBindingTable>&& binding_table,
+      ID3D12DescriptorHeap* heap_for_binding_table,
+      _In_opt_ const DML_BUFFER_BINDING* temporary_resource_binding,
```
adtsai, line 89: There seems to be a slightly inconsistent interface here: some methods take a D3D12BufferRegion (in which case the device context decides which barrier state it needs) whereas others take raw bindings/resources (in which case the caller is the one who has to figure out which state it wants). Same comment goes for the execution context. Should these be refactored to take either one or the other? (Deferring that change to a separate PR is also fine)

tensorflow/core/common_runtime/dml/dml_heap_allocator.cc
jstoecker: README wouldn't be the right place because, to be honest, most people shouldn't be using these variables: they're for development & testing. Good idea to consolidate them in one place though:

https://github.com/microsoft/tensorflow-directml/wiki/Environment-Variables

tensorflow/core/kernels/dml_deepcopy_op.cc
```diff
@@ -42,28 +42,11 @@ class DmlDeepCopyKernel : public OpKernel {
...
-        input_buffer.Offset(), D3D12_RESOURCE_STATE_COPY_SOURCE,
-        output_buffer.SizeInBytes());
-
-    for (auto& barrier : barriers) {
-      std::swap(barrier.Transition.StateBefore, barrier.Transition.StateAfter);
-    }
-
-    execution_context->ResourceBarrier(barriers);
+    uint64_t copy_size =
+        std::min(output_buffer.SizeInBytes(), input_buffer.SizeInBytes());
```
adtsai, line 46: Is there actually a situation where these can differ? Or is this just to be careful?

adtsai (changes requested):

tensorflow/core/common_runtime/dml/dml_device_context.cc
jstoecker: I completely agree, the naming is confusing. I tried to avoid renaming helpers to reduce mental burden of doing this code review (tons of refactoring already) but I'm fine with doing the change here.

tensorflow/core/common_runtime/dml/dml_device_context.h
jstoecker: The goal is to have 95% of the code use D3D12BufferRegion to avoid mistakes. I agree with you on having callers (basically kernels) stick purely with buffer regions and leave the lower-level translation to the execution context. I think I'll make the *Raw helpers on the execution context private and refactor these (Init & Exec) in a separate PR to keep this one from growing too big.

tensorflow/core/kernels/dml_deepcopy_op.cc
jstoecker: Yes! This will trigger in some tests where the input tensor is actually smaller than the output tensor (perhaps padding reasons?). Some other cases where the output is larger than the input. We had some latent bugs that were revealed with this refactoring.

I considered making this logic the default in the device context helper but I'd rather see the CHECK calls fire and understand if there's a bug in a kernel.

tensorflow/core/common_runtime/dml/dml_buffer_region.h
jstoecker: Turns out this isn't possible with a bit more refactoring. Some kernels rely on the buffer regions holding an extra reference on the resource to avoid the allocation (specifically a temporary buffer allocation) from being freed. The buffer regions are currently batched up into the execution context command buffer and keep the resource allocation alive at least until dispatched; changing these to raw pointers means the temporary allocation may be freed immediately once kernel::compute goes out of scope.

The right thing to do would be to refactor the init/exec helpers to take a ref on the allocation but this should be done as a separate change.

tensorflow/core/common_runtime/dml/dml_buffer_region.cc
```diff
@@ -20,21 +20,45 @@ limitations under the License.
...
-  assert(size_in_bytes_ != 0);
+    uint64_t offset, uint64_t size_in_bytes,
+    D3D12_RESOURCE_STATES resource_state,
+    Microsoft::WRL::ComPtr<ID3D12Resource> resource,
+    Microsoft::WRL::ComPtr<ID3D12Resource> resource_copy_src_state,
+    Microsoft::WRL::ComPtr<ID3D12Resource> resource_copy_dst_state)
+    : offset_(offset),
+      size_in_bytes_(size_in_bytes),
+      resource_state_(std::move(resource_state)),
+      resource_(resource),
```
adtsai: ?

adtsai (approved):

tensorflow/core/common_runtime/dml/dml_execution_context.cc
```diff
@@ -109,7 +109,42 @@ DmlGpuEvent DmlExecutionContext::CopyBufferRegion(
   return batch_state_->next_flush_event;
 }
 
-DmlGpuEvent DmlExecutionContext::FillBufferWithPattern(
+DmlGpuEvent DmlExecutionContext::CopyBufferRegion(
```
jeffbloo, line 112: Can D3D12BufferRegion and its callers be simplified by hiding the details of the underlying resource(s)?  

It looks like D3D12BufferRegion could expose methods like ResourceForCopyDst(), ResourceForCopySrc(), and ResourceForUAV() and also have corresponding internal resource pointers and creation parameters, eliminating complexity like below.

tensorflow/core/common_runtime/dml/dml_buffer_region.h
jstoecker: I was mistaken earlier: it wasn't that we needed an extra ref on the ComPtrs, it's that the default move constructor/assignment functions weren't handling the raw pointers and resulted in double freeing temporary buffers.

Issue #294, opened 2021-06-28
Help Needed on Out of Memory Issue


vt911001:
Hi,

I have tried to run SSD MobilenetV1 model from [http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_2018_01_28.tar.gz](http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_2018_01_28.tar.gz) with **tensorflow-directml** precompiled C libraries available in [https://github.com/microsoft/tensorflow-directml/releases/tag/v1.15.5.dev210429](https://github.com/microsoft/tensorflow-directml/releases/tag/v1.15.5.dev210429).

However, I am getting OOM issue when running the model. The snippet of the output is as below:
```
2021-06-28 09:23:32.032871: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2
2021-06-28 09:23:32.035434: I tensorflow/stream_executor/platform/default/dso_loader.cc:99] Successfully opened dynamic library C:\Users\vince\Downloads\Release/directml.adbd007a01a52364381a1c71ebb6fa1b2389c88d.dll
2021-06-28 09:23:32.166394: I tensorflow/core/common_runtime/dml/dml_device_cache.cc:249] DirectML device enumeration: found 1 compatible adapters.
2021-06-28 09:23:32.232103: I tensorflow/core/common_runtime/dml/dml_device_cache.cc:185] DirectML: creating device on adapter 0 (NVIDIA GeForce GTX 950)
2021-06-28 09:23:32.299033: I tensorflow/stream_executor/platform/default/dso_loader.cc:99] Successfully opened dynamic library Kernel32.dll
...
2021-06-28 09:24:26.743588: I tensorflow/core/common_runtime/bfc_allocator.cc:935]      Summary of in-use Chunks by size: 
2021-06-28 09:24:26.743597: I tensorflow/core/common_runtime/bfc_allocator.cc:938] 147330 Chunks of size 256 totalling 35.97MiB
2021-06-28 09:24:26.743604: I tensorflow/core/common_runtime/bfc_allocator.cc:938] 704 Chunks of size 512 totalling 352.0KiB
2021-06-28 09:24:26.743609: I tensorflow/core/common_runtime/bfc_allocator.cc:938] 69 Chunks of size 768 totalling 51.8KiB
[...]

PR #324, opened 2021-08-06
Fix leaks in 'where' and 'tensorarray' kernels

jstoecker:
Normally the op kernel context frees output tensors upon destruction, but calling `release_output()` method will transfer ownership of the tensor object to the caller. Unfortunately, the DML `Where` kernel and `DmlAddToTensor` functions were using this method and never actually freed the allocations. The `Where` kernel leaked three separate allocations:

1. The scalar tensor that holds the count of nonzero coordinates.
2. The 32-bit UINT tensor output from the DML operator (reference to the tensor is overwritten in `ComputeCast` function).
3. The 64-bit INT tensor output from the cast operation.

The `DmlAddToTensor` function is rarely called but leaks the output tensor (presumably to keep to alive until the GPU copy is completed?).

The fix could be cleaner by refactoring some code and using smart pointers, but these are the only places in DML code that call `release_output()` and the refactoring would be riskier than the fix for now.

This should address the leak in #294

### DIFF
--- a/tensorflow/core/kernels/dml_tensor_array.cc
+++ b/tensorflow/core/kernels/dml_tensor_array.cc
@@ -56,8 +56,10 @@ Status DmlAddToTensor(OpKernelContext* ctx, Tensor* sum, const Tensor* current,
 
   TensorValue out = op_ctx.release_output(0);
   ctx->device()->CopyTensorInSameDevice(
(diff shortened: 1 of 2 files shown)


PatriceVignola (approved):

adtsai (approved):
:shipit:

PR #328, opened 2021-08-11
Batch to space / space to batch perf improvements (again)

Jamather:
I added an extra exception for api compatibility when block sizes are not all the same, since DML does not support this. This fixes the test failures associated with the change.


PatriceVignola (approved):

jstoecker (changes requested):
Let's keep this in your private branch for now.

jstoecker (approved):

PR #338, opened 2021-08-23
Link with D3D12 and DXGI/DXCore at runtime instead of load time

jstoecker:
This change removes the load-time dependencies on D3D12, DXGI (Windows), and DXCore (WSL); they are now loaded at runtime like all other dependencies including DirectML. The primary motivation for this change is to make it easy to build the Linux wheels without having access to libd3d12.so and libdxcore.so on the build machine.

### DIFF
--- a/tensorflow/core/common_runtime/dml/dml_util.cc
+++ b/tensorflow/core/common_runtime/dml/dml_util.cc
@@ -21,58 +21,130 @@ limitations under the License.
 
 namespace tensorflow {
 
-Microsoft::WRL::ComPtr<IDMLDevice> TryCreateDmlDevice(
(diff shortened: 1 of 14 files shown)


tensorflow/core/common_runtime/dml/dml_util.cc
```diff
@@ -21,58 +21,130 @@ limitations under the License.
...
-
     return nullptr;
   }
 
-  using DMLCreateDeviceFn = decltype(DMLCreateDevice);
-
-  DMLCreateDeviceFn* dmlCreateDevice;
+  // Load the function.
+  using D3D12CreateDeviceFn = decltype(D3D12CreateDevice);
+  F* create_function;
```
adtsai, line 46: Hmm

adtsai (approved):

PatriceVignola (approved):

PR #340, opened 2021-08-24
Utilize DirectML APIs for BatchNorm and BatchNormGrad in training mode and enable fusion

jeffbloo:
This makes the DirectML backend utilize DirectML APIs for BatchNorm and BatchNormGrad in training mode.  It also enables fusions for BatchNorm operators in this mode.


jstoecker (approved):
Looks good. Just make sure to run clang-format before commit.

PR #341, opened 2021-08-25
Set BFC allocator garbage collection to default false

fdwr:
Avoid garbage collection in the BFC allocator upon running out of memory since chunks/allocations can still be in use on the GPU, which causes a DGXI device removed error and process death.

### DIFF
--- a/tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
+++ b/tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
@@ -57,8 +57,13 @@ bool GPUBFCAllocator::GetGarbageCollectionValue() {
   const char* enable_gpu_garbage_collection =
       std::getenv("TF_ENABLE_GPU_GARBAGE_COLLECTION");
   if (enable_gpu_garbage_collection == nullptr) {
-    // By default, turn on the memory garbage collection.
-    return true;
+    // By default, turn off the memory garbage collection.
+    //
+    // The mainline TF branch sets this to true. This modification
+    // for the DirectML fork is so the garbage collector doesn't
+    // attempt to free chunks in the middle of GPU processing, which
+    // would result in a DXGI device removed error and process death.
+    return false;
   }
   if (strcmp("false", enable_gpu_garbage_collection) == 0) {
     return false;
--- a/tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
+++ b/tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
@@ -53,17 +53,15 @@ bool GPUBFCAllocator::GetAllowGrowthValue(const GPUOptions& gpu_options) {
   return gpu_options.allow_growth();
 }
 
-bool GPUBFCAllocator::GetGarbageCollectionValue() {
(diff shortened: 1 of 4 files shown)


jstoecker (approved):

tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
```diff
@@ -57,8 +57,13 @@ bool GPUBFCAllocator::GetGarbageCollectionValue() {
...
   if (enable_gpu_garbage_collection == nullptr) {
-    // By default, turn on the memory garbage collection.
-    return true;
+    // By default, turn off the memory garbage collection.
+    //
+    // The mainline TF branch sets this to true. This modification for the
+    // DirectML fork is false so the garbage collector doesn't attempt to
+    // free allocations in the middle of GPU processing, which would result
+    // in a DXGI device removed error and process death.
+    return false;
```
PatriceVignola: Can we add a boolean in the constructor (e.g. `default_garbage_collection_value`) that defaults to `true`, and call it with `false` when calling it from the DML BFC allocator constructor? This is not blocking, but even if we won't merge this upstream it would make it easier to port over to TF2.

fdwr: A fine idea. Added.

tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc
```diff
@@ -83,10 +86,11 @@ GPUBFCAllocator::GPUBFCAllocator(SubAllocator* sub_allocator,
 GPUBFCAllocator::GPUBFCAllocator(SubAllocator* sub_allocator,
                                  size_t total_memory,
                                  const GPUOptions& gpu_options,
-                                 const string& name, size_t max_allocation_size)
+                                 const string& name, size_t max_allocation_size,
+                                 bool default_garbage_collection_value)
     : BFCAllocator(sub_allocator, total_memory,
                    GPUBFCAllocator::GetAllowGrowthValue(gpu_options), name,
-                   GPUBFCAllocator::GetGarbageCollectionValue(), 8,
+                   GPUBFCAllocator::GetGarbageCollectionValue(default_garbage_collection_value), 8,
```
PatriceVignola: nit: clang-format

tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.h
```diff
@@ -36,14 +36,15 @@ class GPUBFCAllocator : public BFCAllocator {
...
+                  size_t max_allocation_size = -1,
+                  bool default_garbage_collection_value = true);
   ~GPUBFCAllocator() override {}
 
   TF_DISALLOW_COPY_AND_ASSIGN(GPUBFCAllocator);
 
  private:
   static bool GetAllowGrowthValue(const GPUOptions& gpu_options);
-  static bool GetGarbageCollectionValue();
+  static bool GetGarbageCollectionValue(bool default_garbage_collection_value = true);
```
PatriceVignola: nit: clang-format

PatriceVignola (approved):

PR #343, opened 2021-08-26
Update build pipeline to use 1ES pools and manylinux 

jstoecker:
This change has a few important changes for our CI/release pipeline:

- Linux builds are done using Linux agents (1ES agent pool) instead of Windows agents with a WSL2 distro. This is possible now that the build itself doesn't require WSL interop, and all WSL interop aspects in the YAML pipelines have been removed.
- The actual build step (i.e. build.py) for Linux builds is performed in a manylinux2010-compliant container. This makes our Linux wheels usable in more WSL2 distros (e.g. OpenSUSE Leap 15.2) that couldn't run the current Ubuntu 18.04-built wheels.
- Build.py is updated to produce the C library packages that get uploaded to the TFDML releases (under assets). This was a manual process before, since the bazel target that produces the tarball doesn't handle Windows packages quite right.
- The build pipeline has a runtime parameter to produce non-preview wheels (i.e. no "-dev<date>" in the wheel filename).
- Updated build notes to include Linux steps and notes on manylinux builds.


adtsai (approved):

BUILD.md
```diff
@@ -66,12 +89,26 @@ We've tested the following build environments, and we recommend using a Python e
...
 
 For example, to build the repository and produce a Python wheel use `build.py --package` in a Python 3.5-3.8 environment with `bazel` on the PATH.
 
+# CI/PyPI Builds
+
+The wheels published to [tensorflow-directml PyPI](https://pypi.org/project/tensorflow-directml/) are produced using the Azure Pipelines scripts found under [third_party/dml/ci/pipeline](third_party/dml/ci/pipeline).
+
+# Linux Wheels (Manylinux) and DirectX Libraries
+
+Building *portable* Linux binaries is tricky in comparison to Windows or MacOS. The [Manylinux](https://github.com/pypa/manylinux) project attempts to wrangle some of this complexity by providing standards for build environments to produce binaries that widely usable across GLIBC-based Linux distros. Manylinux also provides container images that implement the various manylinux standards.
```
PatriceVignola: Nit *that are widely usable*

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+
+```bash
+sudo apt install unzip gcc g++
+```
+
+## Install Bazel
+
+```bash
+wget https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh
+bash bazel-0.24.1-installer-linux-x86_64.sh --bin=$HOME/bin/bazel/0.24.1 --base=$HOME/.bazel
```
PatriceVignola: Google officially recommends 0.26.1 for TF 1.15 (and this is the version that we have in the .bazelversion file), so we should probably stick to it. Who knows why they recommend 0.24 for TF 1.14 but explicitly recommend 0.26.1 for TF 1.15.

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+
+Restart your shell if you configured conda to init in the .bashrc file. Otherwise, you can launch the base environment as follows:
+
+```bash
+eval "$($HOME/miniconda3/bin/conda shell.bash hook)" 
+```
+
+After launching the prompt, create a new environment. The examples here use the name `tfdml`, but you can choose any name you like. The idea is to sandbox any packages and dependencies in this environment. This will create a separate copy of Python and its packages.
+
+**IMPORTANT**: make sure to use Python 3.7, 3.6, or 3.5. TensorFlow 1.15 is not supported on newer versions of Python.
```
PatriceVignola: We should remove the 3.5 mention if we decide to stop supporting it.

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+```
+
+Finally, we need to augment the `PATH` of the conda environment to include Bazel.
+
+```
+(tfdml) :~$ mkdir -p $CONDA_PREFIX/etc/conda/activate.d
+(tfdml) :~$ mkdir -p $CONDA_PREFIX/etc/conda/deactivate.d
+(tfdml) :~$ touch $CONDA_PREFIX/etc/conda/activate.d/path.sh
+(tfdml) :~$ touch $CONDA_PREFIX/etc/conda/deactivate.d/path.she
+```
```
PatriceVignola: This is one way to do it, but we should perhaps also suggest the more straightforward approach of using bazelisk as an alternative (non-blocking, we can add it afterwards). I never really liked modifying the activate scripts since I found it to be relatively error-prone.

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+
+Set the contents of the activation script (`%CONDA_PREFIX%\etc\conda\activate.d\path.ps1`). Make sure to use the correct paths where you installed Bazel and MSYS2:
```
PatriceVignola: We should remove this line from the Linux section.

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+
+```bash
+export TFDML_PATH_RESTORE=$PATH
+export PATH=$HOME/bin/bazel/0.24.1:$PATH
+```
+
+Set the contents of the deactivation script (`%CONDA_PREFIX%\etc\conda\deactivate.d\path.ps1`):
```
PatriceVignola: We should also remove this line from the Linux section.

BUILD.md
```diff
@@ -210,3 +247,110 @@ The build script has additional options you can experiment with. To see more det
...
+## Build
+
+Remember to activate your build environment whenever you need to build. To produce the Python package run the following:
+
+```
+(tfdml) :~$ cd ~/src/tensorflow-directml
+(tfdml) :~$ python build.py -p -c release
+```
+
+After the package is built you will find a wheel package under `~/dml_build/python_package`. You can run `pip install` on the output .whl file to install your locally built copy of TensorFlow-DirectML.
```
PatriceVignola: Should we change the default repository to `tfdml_build` instead?

build.py
```diff
@@ -257,6 +261,92 @@ def create_package(args):
...
+      "bazel-bin",
+      "tensorflow",
+      "tools",
+      "lib_package",
+      "libtensorflow.tar.gz")
+
+  bazel_bin_dir = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)),
+    "bazel-bin",
+    "tensorflow")
```
PatriceVignola: Nit: Continuation indentation is 2 here but 4 above (I believe 4 is the right one here).

build.py
```diff
@@ -257,6 +261,92 @@ def create_package(args):
...
+              elif os.path.exists(dml_release_path):
+                  shutil.copy(dml_release_path, dml_target_path)
+              else:
+                  raise "Could not locate DirectML module in redist package"
+
+          # Copy license files.
+          shutil.copy(Path(dml_redist_dir)/'LICENSE.txt', Path(temp_dir)/'DirectML_LICENSE.txt')
+          shutil.copy(Path(dml_redist_dir)/'ThirdPartyNotices.txt', Path(temp_dir)/'DirectML_ThirdPartyNotices.txt')
+
+          shutil.make_archive(Path(args.build_output)/'c_package'/package_name, 'zip', temp_dir)
```
PatriceVignola: This block of code has a different indentation than the rest of the file

build.py
```diff
@@ -257,6 +261,92 @@ def create_package(args):
...
+    "bazel-bin",
+    "tensorflow")
+
+  source_dir = os.path.split(os.path.dirname(os.path.realpath(__file__)))[-1]
+  dml_redist_dir = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)),
+    ("bazel-%s" % source_dir),
+    "external",
+    "dml_redist",
+    "directml")
```
PatriceVignola: This should also have an indentation of 4.

BUILD.md
jstoecker: Let's consider this for a second PR, but it's a good point.

PatriceVignola (approved):

fdwr (approved):

PR #345, opened 2021-08-28
Pin numpy when building

jstoecker:
Fixes issue where python 3.7 builds were using a version of numpy (1.21) that isn't compatible with the version requirement specified by the python wheel (<1.19). We should have left this pinned at 1.18.5 for build purposes, consistent with upstream 1.15.

### DIFF
--- a/third_party/dml/ci/pipeline/setup_agent_windows.yml
+++ b/third_party/dml/ci/pipeline/setup_agent_windows.yml
@@ -38,7 +38,7 @@ steps:
 
 - powershell: |
     Invoke-Expression '$(miniconda.activateCommand)'
-    pip install six numpy wheel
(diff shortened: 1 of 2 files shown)


third_party/dml/ci/pipeline/setup_agent_windows.yml
```diff
@@ -38,7 +38,7 @@ steps:
 
 - powershell: |
     Invoke-Expression '$(miniconda.activateCommand)'
-    pip install six numpy wheel
+    pip install six numpy==1.18.5 wheel
```
PatriceVignola, line 41: We should perhaps have a requirements.txt to avoid duplicating this part (it would also be easier for users to install them all). Shrug.

PatriceVignola (approved):

PR #349, opened 2021-09-16
better logic for forwarding to avoid copies

Jamather:
Allows forwarding in some cases where the output shape does not match the input shape. 
This is an old change which we put off merging until TF GA

### DIFF
--- a/tensorflow/core/kernels/dml_kernel_wrapper.cc
+++ b/tensorflow/core/kernels/dml_kernel_wrapper.cc
@@ -108,22 +108,28 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
     for (int i = 0; i < ctx->num_outputs(); ++i) {
       absl::optional<int> inputIndexToForward =
           shared_helper->GetForwardableInputIndex(ctx, output_shapes, i);
+      
+      bool cancelForward = true;
+
       // If the input is considered forwardable for the op
       if (inputIndexToForward) {
         int inputIndex = inputIndexToForward.value();
-        // Check if the amount of elements is the same.
-        // This ensures output.CopyFrom(input) will never fail.
         const Tensor& input = ctx->input_is_ref(inputIndex)
-                                  ? ctx->mutable_input(inputIndex, false)
-                                  : ctx->input(inputIndex);
+                            ? ctx->mutable_input(inputIndex, false)
+                            : ctx->input(inputIndex);
 
-        if (input.NumElements() == output_shapes[i].num_elements()) {
+        // Element counts must also match
+        if (input.NumElements() == output_shapes[i].num_elements())
+        {
           forwardIndices.emplace_back(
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/dml_kernel_wrapper.cc
```diff
@@ -128,7 +128,14 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
         const Tensor& input = ctx->input_is_ref(inputIndex)
                                   ? ctx->mutable_input(inputIndex, false)
                                   : ctx->input(inputIndex);
-        ctx->set_output(outputIndex, input);
+
+        Tensor output;
+        // Copies underlying data pointer, but uses the output shape provided.
+        if (!output.CopyFrom(input, output_shapes[outputIndex])) {
+          ctx->SetStatus(errors::Internal("Error during copy."));
+        }
```
PatriceVignola:
```suggestion
        OP_REQUIRES(ctx, output.CopyFrom(input, output_shapes[outputIndex]));
```

Just setting the status here won't enough and we will still attempt to call `ctx->set_output(outputIndex, output);` OP_REQUIRES sets the error status but also returns from the function.

tensorflow/core/kernels/dml_kernel_wrapper.cc
```diff
@@ -108,10 +108,19 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
...
-        forwardIndices.emplace_back(
-            std::make_pair(inputIndexToForward.value(), i));
+        int inputIndex = inputIndexToForward.value();
+        // Check if the amount of elements is the same.
+        // This ensures output.CopyFrom(input) will never fail.
+        const Tensor& input = ctx->input_is_ref(inputIndex)
+                                  ? ctx->mutable_input(inputIndex, false)
+                                  : ctx->input(inputIndex);
+
+        if (input.NumElements() == output_shapes[i].num_elements()) {
```
PatriceVignola: What happens to the output if this is false?

tensorflow/core/kernels/dml_kernel_wrapper.cc
```diff
@@ -108,13 +108,28 @@ void DmlKernelWrapperBase::Compute(OpKernelContext* ctx) {
...
+      
+      bool cancelForward = true;
+
+      // If the input is considered forwardable for the op
       if (inputIndexToForward) {
-        forwardIndices.emplace_back(
+        int inputIndex = inputIndexToForward.value();
+        const Tensor& input = ctx->input_is_ref(inputIndex)
+                            ? ctx->mutable_input(inputIndex, false)
+                            : ctx->input(inputIndex);
```
PatriceVignola, line 119: Nit: Indent doesn't look right

PatriceVignola (approved):

PR #391, opened 2022-09-06
Add a small int64 workaround for elementwise ops that don't support it

PatriceVignola:
It turns out that DIVIDE, MODULUS_FLOOR, DIFFERENCE_SQUARE, MODULUS_TRUNCATE and POW don't yet support emulated int64 in DirectML, so we can't entirely get rid of int64 workarounds.

### DIFF
--- a/tensorflow/core/kernels/dml_cwise_ops.cc
+++ b/tensorflow/core/kernels/dml_cwise_ops.cc
@@ -450,8 +450,16 @@ REGISTER_DML_COMPOSITE_BINARY_KERNEL_2(RealDiv, x / y, 8, true, Eigen::half,
 // cwise_op_floor_div.cc).
 REGISTER_DML_COMPOSITE_BINARY_KERNEL_2(FloorDiv, dml::Floor(x / y), 8, true,
                                        Eigen::half, float)
-REGISTER_DML_COMPOSITE_BINARY_KERNEL_3(FloorMod, dml::ModulusFloor(x, y), 8,
-                                       true, Eigen::half, float, int64)
+REGISTER_DML_COMPOSITE_BINARY_KERNEL_2(FloorMod, dml::ModulusFloor(x, y), 8,
+                                       true, Eigen::half, float)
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/dml_cwise_ops.cc
```diff
@@ -481,25 +489,53 @@ REGISTER_DML_COMPOSITE_BINARY_KERNEL_3(Minimum, dml::Min(x, y), 8, true,
...
 // TODO(b/25387198): A special kernel exists for int32 (see cwise_op_add1.cc).
 REGISTER_DML_COMPOSITE_BINARY_KERNEL_3(Add, x + y, 8, true, Eigen::half, float,
                                        int64)
 // TODO(b/25387198): A special kernel exists for int32 (see cwise_op_add1.cc).
 REGISTER_DML_COMPOSITE_BINARY_KERNEL_3(AddV2, x + y, 8, true, Eigen::half,
                                        float, int64)
-REGISTER_DML_COMPOSITE_BINARY_KERNEL_4(TruncateDiv, x / y, 8, true, uint8,
-                                       uint16, int16, int64)
+REGISTER_DML_COMPOSITE_BINARY_KERNEL_3(TruncateDiv, x / y, 8, true, uint8,
+                                       uint16, int16)
```
fdwr, line 519: >uint16, int16 [](http://example.com/codeflow?start=39&length=13)

Should this include uint32 and int32 too? Otherwise it's weird that there's a gap. #Resolved

fdwr (approved):
:shipit:

tensorflow/core/kernels/dml_cwise_ops.cc
PatriceVignola: TruncateDiv doesn't define uint32 for TruncateDiv at the C++ level, and int32 is reserved for CPU.

fdwr: Silly TF.

PR #420, opened 2026-03-21
fix: replace abandoned tf-nightly-2.0-preview with tf-nightly

x4v13r64:
`tf-nightly-2.0-preview` is an abandoned early TensorFlow 2.0 pre-release nightly package. It is no longer maintained or published on PyPI, and the name is unregistered — meaning anyone can publish a malicious package under that name. The `algorithms.ipynb` notebook installs it silently with `pip install -U -q tf-nightly-2.0-preview`, which would resolve and install whatever an attacker registers on PyPI.

This PR replaces it with the current canonical nightly package name `tf-nightly`.

**Security impact:** Any user running the notebook cell would silently install an attacker-controlled package with full access to their environment, credentials, and filesystem.


x4v13r64:
@microsoft-github-policy-service agree

